PR #10, opened 2015-08-17
gen index

jbenet:
This PR adds:
- [x] automated index generation from templates
- [x] list platform download tables based on release bins available
- [x] dist json generation for go-ipfs
- [x] plug in go-ipfs into index gen

Still TODO:
- [x] make "Install Now" button work with js platform detection
- [ ] add ipfs-app dist to index
- [ ] nicer design for index top section + menu


jbenet:
I'm going to merge this. we can PR more later

PR #12, opened 2015-10-20
fix build so it all works on my machine

whyrusleeping:
this makes builds work on my machine


whyrusleeping:
although maybe i misunderstand what publish needs to do. i see now that `index` creates a releases directory

jbenet:
LGTM

PR #14, opened 2015-12-09
[WIP] Front end design & cleanup

Dignifiedquire:
Hopefully closes #11 soon.

Current version:

screen shot 2015-12-09 at 21 28 15

### Todo

- [ ] Design
  - [x] Base styles
  - [ ] Implement scroll spy for navigation
  - [ ] Dividers
  - [ ] Logo 
- [ ] Development
  - [x] Use webpack + harpjs for building
  - [x] Live reloading all the things using browsersync
  - [x] Linting with standard
  - [ ] Pull data from json/markdown files
- [ ] Build step
  - [ ] Integrate `harp compile` into `make`
  - [ ] Integrate `webpack` build into `make`
- [ ] Developer docs

---

You can run this via

```bash
$ npm install
$ npm i -g gulp
$ gulp serve
$ open http://localhost:3000
```

PR #15, opened 2015-12-09
build the binaries on our own

whyrusleeping:
This changes the build code to build the binaries from source ourselves for each platform.


Dignifiedquire:
@whyrusleeping how ready is this? Need to change some of the build process to accommodate for the features in #14

whyrusleeping:
it works for building the go-ipfs binaries (and could be easily extended to other go binaries). but the paths referencing the binary locations will need to be updated. What do you need?

jbenet:
@whyrusleeping rebase over @Dignifiedquire's merged PR?

jbenet:
@whyrusleeping this LGTM.

Dignifiedquire:
@whyrusleeping need to change some things around how the json files are generated and moved.

Dignifiedquire:
@whyrusleeping just checked, looks usable so far, but missing file extensions in the links

whyrusleeping:
Binaries (except on windows) don't have file extensions.

Dignifiedquire:
@whyrusleeping but the file you ship has, e.g. `zip`, `.tar.gz`

Dignifiedquire:
@whyrusleeping what's the current state of this?

whyrusleeping:
@Dignifiedquire this should be pretty good now.

build-go.sh
```diff
@@ -0,0 +1,231 @@
...
+	printf $txtgrn%s$txtnon\\n "$@"
+}
+
+# dep checks
+reqbins="jq zip go"
+for b in $reqbins
+do
+	if ! type $b > /dev/null
+	then
+		fail "must have '$b' installed"
```
jbenet: Install with a makefile and gx (or ipget)? (we can leave this for a future PR)

dists/go-ipfs/Makefile
```diff
@@ -1,32 +1,16 @@
...
-gbcli = $(shell which gobuilder-cli)
 
 all: dist
 
-dist: versions deps
+dist: versions
 	mkdir -p $(releases)
 	cp versions $(releases)/versions
-	./go-ipfs-dist all
+	../../build-go.sh go-ipfs github.com/ipfs/go-ipfs/cmd/ipfs
```
jbenet: pretty sweet! :+1: :smile:

build-go.sh
```diff
@@ -0,0 +1,231 @@
+#!/bin/bash
```
jbenet: this shell file is intense. did you lift it or make it from scratch? 

cc @chriscool for some quick review?

(no need for `sh`, I'm ok to keep `bash` to save time here. all of the people releasing go-ipfs have `bash` installed.)

build-go.sh
```diff
@@ -0,0 +1,231 @@
...
+	fi
+
+	local binname="$1"
+	if [ "$goos" = "windows" ]
+	then
+		binname="$binname.exe"
+	fi
+
+	cp dist.json dist.json.temp
+	jq ".platforms[\"$goos\"].archs[\"$goarch\"] = {\"link\":\"$goos-$goarch/$binname\"}" dist.json.temp > dist.json
```
jbenet: so dist is recreated here?

build-go.sh
```diff
@@ -0,0 +1,231 @@
...
+
+	if [ ! -e $matfile ]; then
+		fail "build matrix $matfile does not exist"
+	fi
+
+	mkdir -p $output
+
+	local distname=$(basename $gobin)
+
+	printInitialDistfile $distname $version > dist.json
```
jbenet: ah ok makes sense :+1:

jbenet:
Ok, this LGTM! :+1: great step fwd

build-go.sh
whyrusleeping: yeah... jq i normally install with npm, zip with my package manager, and go from source. It would be nice to just be able to download them in the script if needed, but this is really just code that we're going to use internally (at least for now)

build-go.sh
whyrusleeping: I wrote this all, lol.

build-go.sh
whyrusleeping: yeah... its kindof a weird hack because doing `cat X | something > X` will leave you with no data.

build-go.sh
Dignifiedquire: Would be nice to have a note about how to install them in the readme. (No idea where and what jq is)

dists/fs-repo-migrations/matrices/v1.0.0
```diff
@@ -0,0 +1,10 @@
+darwin 386
```
Dignifiedquire: Why do we need these? As far as I know 32 bit is not used on any current OS X system that is able to run these binaries (and pretty much nobody, not even [go-lang](https://golang.org/dl/) or [node.js](https://nodejs.org/en/download/) give you 32 bit binaries for OS X)

Dignifiedquire: This applies for ipfs and other binaries as well

Dignifiedquire:
@whyrusleeping I'm afraid not all is working yet, fs-repo-migrations and ipfs-update don't build at all for me, also some of the builds for go-ipfs are failed for me. See the attached gist for details https://gist.github.com/Dignifiedquire/8e8da57025f2377d05c9

dists/fs-repo-migrations/matrices/v1.0.0
whyrusleeping: ah, okay. I just picked a few that seemed reasonable, i'll remove it. But all of these binaries are supported by the go compiler. Which do you think should be kept?

jbenet: Hold on, i think we do need `darwin 386`. I remember running into problems in pre-Yosemite with the mac mini and having to recompile as 386. 

If the go team has it there _still_, it's there for a reason. Let's keep it.

jbenet: (to be clear, i understand that osx has been supposed to run only on `amd64` for some time now, but have seen this specific problem. Until golang disallows `darwin 386`, I think we should continue to have it.

jbenet: (there is also the possibility of suppressing it from the distributions page, if we think "almost all users wont need it and it will be confusing". i twill still be available in the version file listing itself)

build-go.sh
```diff
@@ -0,0 +1,237 @@
...
+}
+
+function checkoutVersion() {
+	local repopath=$1
+	local ref=$2
+
+	if [ -z "$repopath" ]
+	then
+		fail "checkoutVersion: no repo to check out specified"
+	fi
```
chriscool: It's more idiomatic to write something like:

```
test -n "$repopath" || fail "checkoutVersion: no repo to check out specified"
```

build-go.sh
```diff
@@ -0,0 +1,237 @@
...
+
+	echo "==> checking out version $ref in $repopath"
+	cleanRepo $repopath
+	(cd $repopath && git checkout $ref > /dev/null)
```
chriscool: As paths can contain spaces especially on Windows, it's better to quote them like:

```
cleanRepo "$repopath"
(cd "$repopath" && git checkout $ref > /dev/null)
```

build-go.sh
chriscool: Ok, I just commented on a few things that can be applied in many places.

build-go.sh
whyrusleeping: awesome, i knew there had to be a better way

dists/fs-repo-migrations/matrices/v1.0.0
whyrusleeping: mmkay, i left it in.

Dignifiedquire:
Current version viewable at http://v04x.ipfs.io/ipfs/QmP85XXsJoAejv77hcua4mwnkcYUmFAVA2uLmcYU455ttk/index.html

Dignifiedquire:
@whyrusleeping as you can see the platforms is empty for `ipfs-update` though in `dist.json`. Also there is a lot of garbage added when running `make` (things like `gopath` and `dist.json.tmp` that's not cleaned up or gitignored)

Dignifiedquire:
Mobile usable: http://v04x.ipfs.io/ipfs/QmZyvWokPYGg6DrjE6o2V7qhThzZQZ8QCWqdd2U3S75HXC/index.html

Dignifiedquire:
`ipfs-update` is working, after I told my computer to and f*** itself

Dignifiedquire:
Ready for the first round? http://v04x.ipfs.io/ipfs/QmUCqDJ2dVkdfiMrkgsM7PYNVL91GY8LdvpitxKL71biRp/

ralphtheninja:
> Ready for the first round? http://v04x.ipfs.io/ipfs/QmUCqDJ2dVkdfiMrkgsM7PYNVL91GY8LdvpitxKL71biRp/

:tada:

jbenet:
@Dignifiedquire awesome :)

Dignifiedquire:
@jbenet @whyrusleeping can we merge this to master for now? I'd rather do the rest in smaller PRs

Issue #18, opened 2015-12-11
New design pieces


Dignifiedquire:
- [ ] Design
  - [x] Base styles
  - [x] Implement scroll spy for navigation
  - [ ] Dividers
  - [ ] Logo 
- [x] Development
  - [x] Use webpack + harpjs for building
  - [x] Live reloading all the things using browsersync
  - [x] Linting with standard
  - [x] Pull data from json/markdown files
- [x] Build step
  - [x] Integrate `harp compile` into `make`
  - [x] Integrate `webpack` build into `make`
- [ ] Developer docs

PR #19, opened 2015-12-11
[WIP] Front end II

Dignifiedquire:
Solve all the issues in #18


CONTRIBUTING.md
```diff
@@ -0,0 +1,95 @@
...
+* Linting using eslint
+
+
+```bash
+$ gulp
+$ open localhost:3000
+```
+
+
+## How this repo works
```
jbenet: why was this moved here? let's keep all of this in the `README.md` -- the readme of this repo viewed from github should make it very clear how to use the repo. people won't click on `CONTRIBUTING.md` even if we link to it.

site/README.md
```diff
@@ -1,18 +0,0 @@
...
-* JavaScript
-  * bundeling using webpack
-  * transpiling using babel
-* Linting using eslint
-
-
-```bash
-$ gulp
-$ open localhost:3000
-```
```
jbenet: i think it made perfect sense to keep this here. very few people using this repo will need to touch the site. they _should not_ have to know any of this or worry about it.

jbenet: meaning that a user MUST be able to come into this repo, update a version, enter `make` or something close to that, and have everything work. if a user needs to even think about npm, the constraint is not met. `make` should install everything it needs.

tasks/dist.js
```diff
@@ -0,0 +1,24 @@
+const gulp = require('gulp')
```
jbenet: why is gulp in the top level? this should only be used by people messing with the site

dists/go-ipfs/go-ipfs-dist
```diff
@@ -79,8 +79,8 @@ dist_version() {
   cd "$cwd"
 
   # generate dist.json file
-  echo "---> generating dist.json"
-  ./mkdist.js "$version"
+  echo "---> generating config for $version"
+  ../../node_modules/.bin/gulp dist "--path=$(pwd)" "--version=$version"
```
jbenet: no, let's not _force_ users to use gulp. `mkdist.js` worked perfectly fine and only required people having node.

scripts/mkdist.js
```diff
@@ -1,30 +1,30 @@
-var fs = require('fs')
+const fs = require('fs')
+const {join} = require('path')
+const mkdirp = require('mkdirp')
```
jbenet: @Dignifiedquire please don't use this, this forces users to have all the ES2016 harmony stuff just to build the `go-ipfs` dist. ES2016 it's widely deployed yet. this forces installations to match your own, and increases the complexity of messing with this repo.

Across the project, we take "env dependencies" and any "dependencies that need to be installed" very seriously. We don't even use `bash` in _lots_ of places in our scrips and code to ensure that things are more cross-platform and work for users on the first try. It's not easy to do this but it yields much smoother experiences for the large diversity of platforms and dev environments touching our code.

Also, though this `mkdist.js` stuff requires node already, it was (and still is) my intention to remove this dependency and move to shell, or a self-contained go-binary that is installed via `gx` or `ipget`, called by `make`. depending on the users having `node` is a _big_ expectation. depending on them having the shiniest newest babelified harmonified sprinkled with ponies version, is a _huge_ expectation.

i know this is time consuming, and that npm solved so much of this. but most of our repos operate _outside_ of the nice npm bubble. which is why we made `gx` and eventually plan to use it and install it in these repos when necessary.

dists/go-ipfs/mkdist.js
```diff
@@ -1,30 +0,0 @@
-#!/usr/bin/env node
```
jbenet: let's bring this back and remove the `gulp` requirement

jbenet:
@Dignifiedquire most of this LGTM, except for the dependencies problem.

dists/go-ipfs/dist.json
```diff
@@ -2,5 +2,19 @@
...
+      {"id": "openbsd", "name": "OpenBSD App", "browser":"OpenBSD", "icon": "circle-o"},
+      {"id": "freebsd", "name": "FreeBSD App", "browser":"FreeBSD", "icon": "circle-o"},
+      {"id": "windows", "name": "Windows App", "browser":"Windows", "icon": "windows"}
+    ],
+    "archs": [
+      {"id": "386",   "name": "32 bit", "browser": "32"},
+      {"id": "amd64", "name": "64 bit", "browser": "64"},
+      {"id": "arm",   "name": "ARM",    "browser": "ARM"}
+    ]
+  }
```
jbenet: i seem to recall a reason this was in mkdist and _not_ here (i had to move it out after finding a problem) but i don't recall why...

jbenet: one thing to note: not all OSes and platforms listed here will be present in a release. these are just the ones that it _normally_ would be, from which to check. maybe this is why i had it in mkdist, and why the `concat` below is now likely wrong.

dists/go-ipfs/dist.json
```diff
@@ -2,5 +2,19 @@
...
-}
\ No newline at end of file
+  "description": "go-ipfs is the main implementation of IPFS. It is the base distribution, and includes:\n- an IPFS core implementation\n- an IPFS daemon server\n- extensive command line tooling\n- an HTTP API for controlling the node\n- an HTTP Gateway for serving content to HTTP browsers\n",
+  "platforms": {
+    "oses": [
+      {"id": "darwin",  "name": "Mac OSX App", "browser":"OS X",    "icon": "apple"},
+      {"id": "linux",   "name": "Linux App",   "browser":"Linux",   "icon": "linux"},
+      {"id": "openbsd", "name": "OpenBSD App", "browser":"OpenBSD", "icon": "circle-o"},
+      {"id": "freebsd", "name": "FreeBSD App", "browser":"FreeBSD", "icon": "circle-o"},
+      {"id": "windows", "name": "Windows App", "browser":"Windows", "icon": "windows"}
```
jbenet: I don't recall why i called this `<OS> App` -- go-ipfs is not an app, it's a cli toolchain.

CONTRIBUTING.md
Dignifiedquire: This is a work in progress as stated in the title, I moved it out of the way because it needs to be rewritten and updated as it is not up to date anymore. Also details about what tech is used etc, do not belong in the main readme, just quick instructions on how to get things up and running.

site/README.md
Dignifiedquire: I strongly disagree that make files are "nice" they are hard to understand for a lot of people, even I struggle with them everytime I look at them. In addition they require a lot of overhead and understanding if you ever try runnin on something like windows. 

But I don't want to make this pr about a discussion about make vs other build systems. As stated before this is a work in progress and at the end  things will just work by running `make build` as that was what was setup initially. But that will do things like `npm install` as there javascript dependencies in this repo that need to be installed before you can do anything. 

Also one question who is this "user" you are talking about, do you mean yourself and @whyrusleeping (because those are the main people actually using this besides myself right now). Or are you talking about future contributors? In that case see my argument against `make` as that constrains all contributors to be running on unix platforms or do a lot of setup work and we should think about moving everything to be node based instead as well as dropping shell scripts with shebangs which don't work on windows by default either.

tasks/dist.js
Dignifiedquire: All things under `tasks/*` are gulp tasks that's why.

dists/go-ipfs/go-ipfs-dist
Dignifiedquire: Using node means running `npm install` that's all users have to do. And when they have done that gulp is available locally and this works fine without anybody having to wonder about things. 

As for why using gulp, it makes implementing these things easier and I don't have to write the 100th argument parser for a node script

scripts/mkdist.js
Dignifiedquire: Again the user does not need to care about this. This is transpiled using babel when using this through gulp, so all is needed is running `npm install` and things work. 

There is really no point in using node without npm modules. Also the site needs to be build after the releases are built, so without node + npm this repo doesn't work anyway. If you really want to drop the nodejs you will have to find someone else to write this as this means increasing the scope of this project quite large and loosing most attempts of having decent cross platform capabilites.

dists/go-ipfs/mkdist.js
Dignifiedquire: :-1: I'm strongly against these shebangs with node, I have seen to many setups where this simply fail or do not do what people expect them to.

dists/go-ipfs/dist.json
Dignifiedquire: Yes I'm not entirely happy with having to replicate this list in every dist/* folder. But need to see more different dists to find a good solution for this.

dists/go-ipfs/dist.json
Dignifiedquire: How about `<OS> Binary`?

scripts/mkdist.js
jbenet: > If you really want to drop the nodejs you will have to find someone else to write this as this means increasing the scope of this project quite large and loosing most attempts of having decent cross platform capabilites.

i don't intend to blow up scope at all, i want to keep it even smaller than the current PR. 

The mkdist changes aren't necessary here. I do intend to get rid of node as a requirement _in the long run_ (though not now and not asking you to). 

I also do not mean to make it harder to work with this repo than it needs to be-- which is why I made this using node the very first time: a quick + dirty solution that worked for me + david. And why i'm fine with ingesting the new node requirements, babel, and so on for the site. But mkdist does not need to ingest more dependencies.

I really appreciate you taking the time to clean other things up -- thank you! 

But the extra changes are beyond scope here -- as mentioned in last PR, all i asked for is changes to the _display_ of the site. (as an extreme, i would have expected the changes to be doable in css alone). Other changes happened, and that's very fine because it's what you prefer and the site looks way, way better now, etc. (thanks!!). But im not asking for mkdist and the rest of the repo to change.

---

I write the below to make it more clear _why_ i'm **against requiring installing node** in general  -- as much as I am against **requiring installing go or other large dev envs when not necessary**. I don't mean to turn this into a long discussion either -- only to express why i may be seemingly-unreasonably rejecting what _i think_ is the best package manager + smooth package install process created to date.

node + npm are certainly very nice, but you have to buy into installing node + npm which is _not_ that nice. then you have to keep them up to date, have to learn how to use new things, and debug them when they go wrong (which they inevitably do, often. even in npm deps get out of date and break). It's hard for regular node users to look back and realize just how hard it is to get going with node and installing things in the first place. it's not a "30 seconds and done forever" thing, it's something that introduces knowledge dependencies and forces the user to learn more things, inc some very complicated things.

Look, it's the same story with go. `go get` + compiling things is really, really nice. everything "just works", it's cross platform, everything is awesome. That is, _if_ you buy into installing go, setting up your env + path the right way, and you keep your packages nice and not breaking interfaces. By the time you get to installing the package, the user has had to learn a ton of things. (both Go and Node are similar in this, and why putting node/npm and golang people in the same room usually ends with people talking past each other about how their thing is the best without trying each other's stuff out. (fun for the reader: have rob pike and substack write programs in js and go respectively. hilarity ensues.)

Both node/npm and go models are really nice, if you buy into them fully, have them installed, and know how to use them. But the problem is it makes repos unapproachable to people outside of those circles, people who do not yet have these things installed, and who honestly should not need to. I've watched hundreds of people install Go and node just to use a small script or program. It's very painful.

dists/go-ipfs/mkdist.js
whyrusleeping: yeah, this broke on two of my machines because my binary (from my package manager) was called `nodejs`

jbenet: im not talking about this shebang, i'm talking about not forcing gulp for everything. I'm taling about bringing back _this file_.

---

Fwiw, 
- the shebang is standard and tons of modules use it, see: https://github.com/search?q=%22%23%21%2Fusr%2Fbin%2Fenv+node%22&ref=searchresults&type=Code&utf8=%E2%9C%93
- but the shebangs are _known to break on windows_, so it's not properly cross platform.
- but having node as `nodejs` is orthogonal to this. 
- and a bin called `nodejs` is going to break with the way that the entire node ecosystem calls node, from packages and from npm `package.json` scripts (by calling `node $args` directly). 
- beyond this, should probably file a bug with arch linux. i'm sure they have reasons to call it `nodejs` but it does break tons of shit. :(

dists/go-ipfs/go-ipfs-dist
```diff
@@ -80,7 +80,7 @@ dist_version() {
 
   # generate dist.json file
   echo "---> generating dist.json"
-  ./mkdist.js "$version"
+  `which node` ./mkdist.js "$version"
```
Dignifiedquire: I know this doesn't solve the issues, but just want to pull it out, and make clear it's js file getting executed via node

Dignifiedquire:
All merged into #15

Issue #24, opened 2016-01-13
user testing feedback


jbenet:
@Dignifiedquire this is looking great -- took it out for a spin with people.

some feedback:

- i ran this by 6 people. 3 were hackers. 1 other had coded something at some point. 2 were non technical.
- they all generally liked the page ("looks {good, ok, fine, functional} to me")
- they all were fine navigating
- they all were fine with the columns + clicking.
- in iOS safari it rendered weird, the columns didnt go away after click, and the anchor was a bit offset.
- 2 had trouble noticing that "docs / changelog / all versions" was clickable. 
- it is not very clear that blue text is links in this page. it's a different color and the glow isn't very noticeable in some screens. an underline on hover would be good. 
- 4 had trouble knowing where to download. didnt know the platform grid was links-- (1 thought it was just a listing of what it worked with). "Ok, how do i download it?" was a common question.
- 3 people suggested ["a (big) (green) button with a down arrow"](https://ipfs.io/ipfs/QmZ5hMju6mEAD8LyoBtP5ZFGbio3mjSE5zhqDbUvhuajnu/cap.png).
- might be ok to add "Downloads:" as a title above the grid (i dont like how it looks, but it makes it clear. -- or "Choose platform to download:")
- but what will be very good is to have smart button that detects user's OS and downloads that platform. fine to delay till later if much trouble.
- people were distracted by much text-- fixed when i made the min-size of each distribution "page" (`.d-component { min-height: 700px }` -- about a screen-height's worth. when we add the asciinema or jpgs and so on it will get there, but may be good to just do it (even it blank) to make sure the user is focused on one release at a time even if there is no media.
- (surprising to me) people were mostly fine once in the "all versions" part. -- crude dir listing was ok. they got to the zip! but they did snag on the platforms listing 
- 2 people were looking for osx and did not know "what the hell is darwin?" so we should consider renaming from Go's to more common OS name

![](https://ipfs.io/ipfs/QmPodAjcN6HuT41Q1ZFyEwpz2rdkHEz7kc6rCRiWMQbNaw/cap.png)
![](https://ipfs.io/ipfs/QmNbLfzJhYmNZCF6VyTTV4mJTKfp4yUVgM8BF3gscMkjvs/cap.png)

i can do more user testing if useful.

jbenet:
oh:
- 2 people wanted to go to ipfs homepage from here. they tried clicking "ipfs" in the top left. or wanted more links in the top bar, or "as the first entry in the navigation"

[...]

PR #26, opened 2016-01-14
Improve on the UI

Dignifiedquire:
Start resolving the issues from #24 

- [x] Bit fat download button with os detection
- [x] Increase hight
- [ ] Add underlines to links
- [ ] Fix mobile version issues

Current preview

screen shot 2016-01-14 at 22 06 33

@jbenet please take a look

@whyrusleeping 0.3.11 is not setup for being built yet could you add that please?


jbenet:
Button + height LGTM!

jbenet:
(button looks great actually)

jbenet:
just thought: should we add OS icons from font-awesome wherever relevant? (say, to left of the text in the grid, and to the auto-detect button?)

jbenet:
could we use a down-arrow on the button? common symbols help catch attention.

Dignifiedquire:
> could we use a down-arrow on the button? common symbols help catch attention.

I'd rather not add an icon to the download button, as I already needed to make the font quite small as some of those names are very long (I'm looking at you `fs-repo-migrations`) and there is simply not enough space on some screen sizes to add it

Dignifiedquire:
> just thought: should we add OS icons from font-awesome wherever relevant? (say, to left of the text in the grid, and to the auto-detect button?)

I think it's not needed on the download button, will think about adding it to the grid, but I haven't yet found a high quality free bsd icon that works in small sizes to use, so I'd rather not add them if we don't have good ones for all platforms.

> (button looks great actually)

Thanks :)

RichardLitt:
Code LGTM

jbenet:
@Dignifiedquire 
- what about just the icon instead of repeating the name? or just the icon when the screen width is below some threshold  (`@media` queries, woo)?
- arrows matter, so does green. Green + down arrow are both action buttons across tons of sites and people's minds are well conditioned. (i once ran an a/b test with green button vs gray button and green was ~5x more likely to be clicked.) i know it fits less with the blue, _any_ shred of confusion finding the button is worse than a feeling the colors don't entirely fit.

jbenet:
- we should add the icons at some point, but i'm ok delaying it for a future PR. 
- "speed of finding download link and clicking it" is way more important than "the icons look a bit bad" here. 
- freebsd: https://openclipart.org/image/2400px/svg_to_png/226471/1441081059.png
- maybe the puffer to the right for openbsd?

Dignifiedquire:
> arrows matter, so does green. Green + down arrow are both action buttons across tons of sites and people's minds are well conditioned. (i once ran an a/b test with green button vs gray button and green was ~5x more likely to be clicked.) i know it fits less with the blue, any shred of confusion finding the button is worse than a feeling the colors don't entirely fit.

Okay two things

* Down arrow. I'm not sure why you think this is so important, I know it was used for some time but these days I have a hard time to find any large download page that still uses an arrow for the download button. (See below for some examples of popular download pages that just manage fine without icons)
* Green vs Grey. The reason for your a/b test to go to green is clear, but not because of green, but rather because it was colored (green) vs uncolored (gray) so I see no conclusion from that to use a green button here in a design that was not at all intended to use green. If you want a different highlight color we *could* use orange, as it actually fits and contrasts nicely to blue, but I'm not sure that kind of highlight is needed

### Examples

#### Chrome
screen shot 2016-01-15 at 10 03 41

#### Adobe
screen shot 2016-01-15 at 10 03 51

#### Go-lang
screen shot 2016-01-15 at 10 04 46

Dignifiedquire:
> "speed of finding download link and clicking it" is way more important than "the icons look a bit bad" here.

Speed of finding the link is important I totally agree, but if the icons are not actually distinguishable and look bad they actually have the opposite effect as people are then spending their time trying to figure out what the icon is supposed to mean.

Dignifiedquire:
New version

Issue #33, opened 2016-01-20
Go downloads hierarchy is broken


jbenet:
The version listings for the go packages got screwed up, they now have a _directory_ for a version with just a `go-ipfs.zip` inside. This is broken, please restore the old schema, which is typical of traditional systems, and which tools already depend on:

- `go-ipfs_$version_$platform.$archive` -- eg `go-ipfs_v0.3.11_amd-64.tar.gz` - this way it is clear on a downloads folder what version you just downloaded, and it does not replace older versions.
- `.zip` is fine for windows, but please use `.tar.gz` for unix platforms, as some zip tools ignore/lose permissions. since we will ship binaries and shell scripts it is important that they _absolutely always_ keep permissions.

To be clear, we want: 

```
.
├── go-ipfs
│   ├── latest -> v0.3.7
│   ├── v0.3.6
│   │   ├── go-ipfs_v0.3.6_darwin-386.tar.gz
│   │   ├── go-ipfs_v0.3.6_darwin-amd64.tar.gz
│   │   ├── go-ipfs_v0.3.6_linux-386.tar.gz
│   │   ├── go-ipfs_v0.3.6_linux-amd64.tar.gz
```

jbenet:
this is important before formally shipping it in a blog post, or writing tools for it

jbenet:
Please respect design constraints as speced out in https://github.com/ipfs/distributions#how-this-repo-works

dignifiedquire:
This is fixed in the latest version

PR #38, opened 2016-01-21
fix(build): Change package name to match the constraints

Dignifiedquire:
This fixes the package names, but is still missing the switch to `.tar.gz` for some oses.

Ref #33

### DIFF
--- a/build-go.sh
+++ b/build-go.sh
@@ -50,14 +50,18 @@ function doBuild() {
 	local goarch=$2
 	local target=$3
 	local output=$4
+  local version=$5
+
+  dir=$output
+	name=$(echo $target | awk -F'/' '{ print $3 }')
+  binname=${name}_${version}_${goos}-${goarch}
 
 	echo "==> building for $goos $goarch"
 
-	dir=$output/$1-$2
-	if [ -e $dir ]; then
-		echo "    $dir exists, skipping build"
-		return
-	fi
+	# if [ -e $dir ]; then
+	# 	echo "    $dir exists, skipping build"
+	# 	return
+	# fi
 	echo "    output to $dir"
 
 	mkdir -p tmp-build
@@ -76,7 +80,6 @@ function doBuild() {
 	fi
 
 	# now zip it all up
-	binname=$(echo $target | awk -F'/' '{ print $3 }')
 	mkdir -p $dir
 	if  zip -r $dir/$binname.zip tmp-build/* > /dev/null; then
 		printDistInfo $binname
@@ -130,7 +133,7 @@ function buildWithMatrix() {
 	# build each os/arch combo
 	while read line
 	do
-		doBuild $line $gobin $output
+		doBuild $line $gobin $output $version
 	done < $matfile
 
 	mv dist.json $output/dist.json
--- a/build-go.sh
+++ b/build-go.sh
@@ -50,11 +50,11 @@ function doBuild() {
 	local goarch=$2
 	local target=$3
 	local output=$4
-  local version=$5
+	local version=$5
 
-  dir=$output
+	dir=$output
 	name=$(echo $target | awk -F'/' '{ print $3 }')
-  binname=${name}_${version}_${goos}-${goarch}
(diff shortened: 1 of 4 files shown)


whyrusleeping:
do we really need to switch to .tag.gz for some? thats gonna be annoying since we have to hard code that  in per platform

build-go.sh
```diff
@@ -50,14 +50,18 @@ function doBuild() {
 	local goarch=$2
 	local target=$3
 	local output=$4
+  local version=$5
```
whyrusleeping: indenting here is off

build-go.sh
```diff
@@ -50,14 +50,18 @@ function doBuild() {
...
 
 	echo "==> building for $goos $goarch"
 
-	dir=$output/$1-$2
-	if [ -e $dir ]; then
-		echo "    $dir exists, skipping build"
-		return
-	fi
+	# if [ -e $dir ]; then
+	# 	echo "    $dir exists, skipping build"
```
whyrusleeping: mmm, if we're going to force a rebuild every time, maybe just have a flag that deletes the dir. I don't like the idea of building in dirty dirs

build-go.sh
```diff
@@ -130,7 +133,7 @@ function buildWithMatrix() {
 	# build each os/arch combo
 	while read line
 	do
-		doBuild $line $gobin $output
```
whyrusleeping: where is this version variable coming from?

build-go.sh
Dignifiedquire: I don't think we should force it, but the issue is that this needs to be either moved so that we only check it once per version (not per build as at the moment)  or we need to change the check to check folder + target file.

build-go.sh
Dignifiedquire: stupid tabs..

Dignifiedquire:
> do we really need to switch to .tag.gz for some? thats gonna be annoying since we have to hard code that in per platform

See #33 for @jbenet argument

build-go.sh
Dignifiedquire: No idea but it's there, and you used it on line 130 :laughing:

whyrusleeping: i used it on line 130 because i declared it in that function. it wasnt passed into this one.

Dignifiedquire: I'm confused, l 130 is part of this function no?

Dignifiedquire: Disclaimer: I have no idea about bash so maybe I'm missing sth obvious.

Dignifiedquire:
New version with from this branch http://v04x.ipfs.io/ipfs/QmUA7DPsDZrp1qjDHfG9oGZM7Z1Jx7fosawHxu7B78rHz3/ @jbenet @whyrusleeping please review

build-go.sh
jbenet: lines: 130 in old, 165 in new.

yes its part of this function. i think this is fine.

jbenet:
thank you @Dignifiedquire this LGTM.

jbenet:
@Dignifiedquire i think this snuck in?

- http://v04x.ipfs.io/ipfs/QmUA7DPsDZrp1qjDHfG9oGZM7Z1Jx7fosawHxu7B78rHz3/go-ipfs/v0.3.11/build-log (shows build errors)

Dignifiedquire:
@jbenet yes looks like it

jbenet:
@whyrusleeping sorry-- tar/gz is much more reliable for these things. among problems, zip is not guaranteed to preserve permissions. some zip tools dont respect them.  also AFAIK tar is also much more present, if we want to get ipfs into an embedded system, not having to install zip first is nice (tar is much more likely to be there).  (( believe me i like zip more, its easier to reason about, but tar is AFAIK safer ))

Dignifiedquire:
@jbenet moved here: https://github.com/ipfs/distributions/issues/44

whyrusleeping:
@jbenet oh no, I like tar WAYYYYYY more than zip. But i just hate having to do two different things. its ugly

PR #42, opened 2016-01-21
adds 'ipget' to distributions

noffle:
My first swing at this. Let me know if I've missed anything.

The build is currently failing for `freebsd 386` and `freebsd arm`, and also all arches of Windows. I'll investigate these separately.


jbenet:
@noffle can you include a link to a build of the site?

noffle:
I'm having trouble with the gateways right now, so here it is (sans actual binaries) on my site: http://www.stephenwhitmore.com/tmp/releases/

I realized in building it that the site makes some deep assumptions about all repos being under the `github.com/ipfs/` root. I'll address this in another PR.

whyrusleeping:
weird thing, the progress meter on the left never gets to ipget, when scrolled all the way down for me its still on the previous one

noffle:
ping @dignifiedquire @whyrusleeping

whyrusleeping:
LGTM

dignifiedquire:
What about the site and changelog links? At the moment they will not work, so I suggest we either make it generic so that each product can bring its own links or we move ipget to the ipfs organization on github for now.

noffle:
@dignifiedquire: good call -- done.

jbenet:
Looks like this needs rebasing

dignifiedquire:
@noffle a couple of things:

* The links do not work currently: They look like this: `https://github.com/ipfs/noffle/tree/ipget/README.md`
* The dist files have `diff` in the middle Oo, e.g `ipget_diff_darwin-386.tar.gz`

noffle:
Sorry for the noise @dignifiedquire. Addressing these.

> The links do not work currently: They look like this: https://github.com/ipfs/noffle/tree/ipget/README.md

Bad merge; fixed.

> The dist files have diff in the middle Oo, e.g ipget_diff_darwin-386.tar.gz

I looked into this. Apparently `build-go.sh` uses `git show --pretty="%H"` to get the current hash. However, `git show` outputs in [combined diff](http://git-scm.com/docs/diff-format) format, which will show a diff when there are files that differ between sides. This means the "sha output" put into the script ended up being:

```
% git show --pretty="%H"
dbc4e76c354276c5666005c587b40379bb382093

diff --git a/dists/ipget/mkdist.js b/dists/ipget/mkdist.js
new file mode 100644
index 0000000..abb6999
--- /dev/null
+++ b/dists/ipget/mkdist.js
@@ -0,0 +1,28 @@
+var mkdist = require('../../scripts/mkdist')
+
+if (process.argv.length != 3) {
+  console.log('usage: mkdist.js <version>')
+  process.exit(1)
+}
...
```

I've included another commit here to fix this issue in `build-go.sh`.

dignifiedquire:
Thanks @noffle merged

PR #51, opened 2016-02-08
feat: Setup code signing for all packages using opgenpgp

dignifiedquire:
**Update**: Added process references
**Update**: Added sha512 checksums

I've setup a pgp on my yubikey that is used for the code signing process. This process generates pgp signatures for every single tar/zip file and adds them to directory. As in the description explained they can be verified by running 

``` bash
$ gpg --verify file.tar.gz.asc file.tar.gz
```

if the key is present. 

Ref https://github.com/ipfs/go-ipfs/issues/957
## References
- The distribution of keys is modeled after https://developers.yubico.com/Software_Projects/Software_Signing.html 
- The signature generation is modeled after the release signing process of the apache foundation: https://www.apache.org/dev/release-signing.html
### Setup
- Every developer (currently me) who has the right to sign releases generates a pgp key, stored on dedicated hardware like a yubikey
- The fingerprint is added to dist.ipfs.io
- The public key is uploaded to https://pgp.mit.edu/ for ease of distribution
### Verification
- The user who downloads from dist.ipfs.io can fetch the listed public keys listed from https://pgp.mit.edu/ and add them to their keyring
- The user downloads the signature + file
- The user uses pgp + signature to verify the file 
## Checksums

In addition to the signatures there are also `file.sha` files containing the `sha512` checksum of the packages.

### DIFF
--- a/SIGNING.md
+++ b/SIGNING.md
@@ -1,0 +1,56 @@
+# Signing and Security
+
+In order to ensure that downloaded binaries are not compromised we provide
+two ways of checking the integrity of the downloaded files.
+
+In the following the reference to "tarball" means either a `zip` or `tar.gz` file
+depending on the target operating system.
+If not stated otherwise "key" refers to a public/private key pair usable for public
+key cryptography.
+
+## 1. `SHA512` Checksum
+
+After the tarball was created, [`gpg`](https://gnupg.org/) is used to generate
+the `SHA512` checksum of the tarball and put into a file called `$original_file.sha`.
+The command for doing this is
+
+```bash
+$ gpg --print-md SHA512 $original_file> > $original_file.sha
+```
+
+## 2. OpenPGP Compatible ASCII Armored Detached Signatures
+
+After the tarball is generated it is signed using one of the trusted developer keys.
+From that an OpenPGP compatible ASCII armored detached signature is created and
+put into a file `$original_file.asc`.
+The command for doing this is
+
+```bash
+$ gpg --armor --output $original_file.asc --detach-sig $original_file
+```
+
+## Trusted Developer Keys
+
+These keys are the ones used to sign tarballs and used to verify their integrity.
+
+### Required properties of the keys
+
+* The private key MUST be stored on seperate hardware than the computer used to sign
+  the release. For convenience something like a [YubiKey](https://www.yubico.com/)
+  is recommended.
+* The key must have a length of at least `2048` bits and of type RSA.
+* The public key MUST be uploaded to https://pgp.mit.edu/.
+* The full fingerprint MUST be listed on the distributions page.
+
+### Obtaining the public keys for verification
+
+The keys fingerprints are listed on the distributions page and the public keys
+can be downloaded from https://pgp.mit.edu/ using
+
+```bash
+$ gpg --keyserver pgpkeys.mit.edu --recv-key $figerprint
+```
+
+## Further Reading
+
+* [Apache Release Signing Document](https://www.apache.org/dev/release-signing.html)
--- a/site/public/_about.md
+++ b/site/public/_about.md
@@ -27,10 +27,13 @@ $ gpg --verify go-ipfs.tar.gz.asc go-ipfs.tar.gz
 
 You will need to download the public key of the release managers, which are currently,
 
-* Friedel Ziegelmayer <dignifiedquire@gmail.com> [`27F50659`](https://pgp.mit.edu/pks/lookup?search=0x27F50659&op=vindex&fingerprint=on).
+* Friedel Ziegelmayer <dignifiedquire@gmail.com> [`E2C5 3DFE 7CBA 9864 38B9  88D9 0741 3B8A 27F5 0659`](https://pgp.mit.edu/pks/lookup?search=0xE2C53DFE7CBA986438B988D907413B8A27F50659&op=index&fingerprint=on&exact=on).
 
 The command for this is
 
 ```bash
 $ gpg --keyserver pgpkeys.mit.edu --recv-key <keyid>
 ```
+
+There is also a `file.sha` file for every package that contains the `sha512` checksum
+as an additional integrity check.
--- a/site/public/_about.md
+++ b/site/public/_about.md
@@ -16,3 +16,21 @@ page with,
 The `All Versions` link on each distribution shows directory listings for all the available versions, and a `versions` file ([example](http://dist.ipfs.io/go-ipfs/versions)). This file can be used by tools, such as [ipfs-update](#ipfs-update), to find all the available versions and download the latest.
 
 The directory listing of each version ([example](http://dist.ipfs.io/go-ipfs/v0.3.11)) has all the platform archives (`.zip` or `.tar.gz`), a `README.md` and a `dist.json` which describe the release for humans and machines. It is meant to be easily consumed and used by tools.
+
+##### Code Signing
+
+All releases are signed using [OpenPGP](http://www.openpgp.org/). You can verify this by running
+
+```bash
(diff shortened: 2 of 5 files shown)


dignifiedquire:
cc @jbenet @whyrusleeping

site/public/_about.md
```diff
@@ -16,3 +16,21 @@ page with,
...
+
+All releases are signed using [OpenPGP](http://www.openpgp.org/). You can verify this by running
+
+```bash
+$ gpg --verify go-ipfs.tar.gz.asc go-ipfs.tar.gz
+```
+
+You will need to download the public key of the release managers, which are currently,
+
+* Friedel Ziegelmayer <dignifiedquire@gmail.com> [`27F50659`](https://pgp.mit.edu/pks/lookup?search=0x27F50659&op=vindex&fingerprint=on).
```
Kubuxu: It would be better to have full fingerprint here.

dignifiedquire:
sure, curious why are they not good anymore?

Kubuxu:
There is a list of all short fingerprints generated, if someone will not look at the name and email there is a chance he will download a wrong one from the keyserver.

dignifiedquire:
thanks @Kubuxu will update with this in mind

jbenet:
Would be nice to have multiple sigs here (like say @whyrusleeping and i both signing off to a release)

jbenet:
@dignifiedquire thank you for leading the charge here! :+1: :heart:

dignifiedquire:
@jbenet yes need more sigs for sure, but I don't have yours lying around ;) So @jbenet @whyrusleeping please make a PR adding your sigs in here, preferably from a key on your yubikey

whyrusleeping:
guide: https://www.yubico.com/2012/12/yubikey-neo-openpgp/

dignifiedquire:
TODO from my side, write up general strategy for review

dignifiedquire:
@jbenet added first draft of a security doc describing the different parts.

SIGNING.md
```diff
@@ -0,0 +1,56 @@
...
+## Trusted Developer Keys
+
+These keys are the ones used to sign tarballs and used to verify their integrity.
+
+### Required properties of the keys
+
+* The private key MUST be stored on seperate hardware than the computer used to sign
+  the release. For convenience something like a [YubiKey](https://www.yubico.com/)
+  is recommended.
+* The key must have a length of at least `2048` bits and of type RSA.
```
dignifiedquire: I know that generally keys above `4096` should be used, but right now I haven't found a way to generate those keys on my yubikey. So if someone knows how please let me know and we can increase this.

SIGNING.md
```diff
@@ -0,0 +1,56 @@
...
+key cryptography.
+
+## 1. `SHA512` Checksum
+
+After the tarball was created, [`gpg`](https://gnupg.org/) is used to generate
+the `SHA512` checksum of the tarball and put into a file called `$original_file.sha`.
+The command for doing this is
+
+```bash
+$ gpg --print-md SHA512 $original_file> > $original_file.sha
```
Kubuxu: How about quoting `$original_file` and `original_file.sha`?
I know it is only sample but it is good practice.

SIGNING.md
Kubuxu: Only Yubikey 4 and 4 Nano support `RSA 4096`.

site/public/_about.md
```diff
@@ -16,3 +16,24 @@ page with,
...
+
+You will need to download the public key of the release managers, which are currently,
+
+* Friedel Ziegelmayer <dignifiedquire@gmail.com> [`E2C5 3DFE 7CBA 9864 38B9  88D9 0741 3B8A 27F5 0659`](https://pgp.mit.edu/pks/lookup?search=0xE2C53DFE7CBA986438B988D907413B8A27F50659&op=index&fingerprint=on&exact=on).
+
+The command for this is
+
+```bash
+$ gpg --keyserver pgpkeys.mit.edu --recv-key <keyid>
+```
```
Kubuxu: Instead of doing this, which means that user has to import the key and will get warning that files are signed by key he doesn't trust. We could use `gpgv`/`gpgv2` and custom keyring file. See: http://unix.stackexchange.com/a/78476

dignifiedquire: So we would distribute a keyring file which includes all the trusted keys? How would we best do that?

Kubuxu: Including keyring/IPFS hash with previous distribution wouldn't be that bad of an idea.

Other option is just direct /ipfs/ link.

SIGNING.md
```diff
@@ -0,0 +1,56 @@
+# Signing and Security
+
+In order to ensure that downloaded binaries are not compromised we provide
+two ways of checking the integrity of the downloaded files.
+
+In the following the reference to "tarball" means either a `zip` or `tar.gz` file
+depending on the target operating system.
+If not stated otherwise "key" refers to a public/private key pair usable for public
```
RichardLitt: If not stated otherwise**,** "key

SIGNING.md
```diff
@@ -0,0 +1,56 @@
...
+key cryptography.
+
+## 1. `SHA512` Checksum
+
+After the tarball was created, [`gpg`](https://gnupg.org/) is used to generate
+the `SHA512` checksum of the tarball and put into a file called `$original_file.sha`.
```
RichardLitt: and **then it is** put into

RichardLitt:
LGTM.

SIGNING.md
```diff
@@ -0,0 +1,56 @@
+# Signing and Security
+
+In order to ensure that downloaded binaries are not compromised we provide
+two ways of checking the integrity of the downloaded files.
+
+In the following the reference to "tarball" means either a `zip` or `tar.gz` file
```
jbenet: use "archive"

SIGNING.md
jbenet: - use `.sha256` extension
- Why not use the `multihash` tool? that way we can ratchet it up.

SIGNING.md
```diff
@@ -0,0 +1,56 @@
...
+These keys are the ones used to sign tarballs and used to verify their integrity.
+
+### Required properties of the keys
+
+* The private key MUST be stored on seperate hardware than the computer used to sign
+  the release. For convenience something like a [YubiKey](https://www.yubico.com/)
+  is recommended.
+* The key must have a length of at least `2048` bits and of type RSA.
+* The public key MUST be uploaded to https://pgp.mit.edu/.
+* The full fingerprint MUST be listed on the distributions page.
```
jbenet: - it is good to allow users to verify using standard procedures.
- it would be nice to provide our own route with our own tools. put the keys themselves in the distributions page, distribute the public keys along in the git repo and in the distributions ipfs dag. both are merkle-dags, if there's changes / conflicts, we've got more serious issues.

SIGNING.md
Kubuxu: We use Neo which supports only 2048.

djdv:
Recently, there has been discussion in other repos about creating source tarballs that could be built offline due to various connection issues (lost hash, geographic censorship).
https://github.com/ipfs/ipget/issues/48#issuecomment-365267198
https://github.com/ipfs/go-ipfs/issues/4765#issuecomment-370406373

These releases will have to be signed for user saftey if they are to be ferried by third parties across various boundaries. As such, I simply want to pulse an alert on this issue due to its age and significance.
@whyrusleeping @Kubuxu

hsanjuan:
@Stebalien should we pick this up?

Stebalien:
That would be awesome! Note: I think a higher priority is a way to update dists without having to fetch and re-add everything if you're looking to improve dists.

Stebalien:
Closing as stale.

PR #65, opened 2016-05-18
Enable go-ipfs builds for NetBSD

lgierth:


lgierth:
I refrained from adding a NetBSD build to the existing 0.4.2 release because effort :) Next release will include NetBSD.

mburns:
Is this something we still want to land? What work is required?

Stebalien:
go-ipfs doesn't currently build on netbsd (fuse is broken)

PR #75, opened 2016-08-11
Cleanup and update all dependencies

dignifiedquire:
Makes it work with node@6

cc @whyrusleeping


whyrusleeping:
LGTM (as much as all this weird node stuff can look good to me)

Issue #133, opened 2017-01-31
Add orbit distributions


RichardLitt:
This issue tracks the relevant steps needed to add Orbit to dist.ipfs.io, specifically referring to code which needs to happen on this branch. See https://github.com/haadcode/orbit/issues/132 for an overarching issue about how to make this happen. 

Current steps needed AFAIK:
- [ ] Create build-js.sh
- [ ] Edit makefile for Orbit to be able to build this specifically
- [ ] Figure out which versions ([npm releases](https://github.com/haadcode/orbit/releases)?)
- [ ] Merge PR
- [ ] Check

RichardLitt:
@haadcode Which releases do we want on this?

RichardLitt:
@haadcode I ran into a problem with building orbit, not sure why. I'm still getting the old bug:

```
  inflating: v3.2.0/orbit-3.2.0/test/orbit.test.js
--> making v3.2.0
npm ERR! git fetch -a origin (https://github.com/haadcode/uport-persona-1.git) remote: Repository not found.
npm ERR! git fetch -a origin (https://github.com/haadcode/uport-persona-1.git) fatal: repository 'https://github.com/haadcode/uport-persona-1.git/' not found
npm ERR! Darwin 15.6.0
npm ERR! argv "/Users/richard/.nvm/versions/node/v6.4.0/bin/node" "/Users/richard/.nvm/versions/node/v6.4.0/bin/npm" "i"
npm ERR! node v6.4.0
npm ERR! npm  v3.10.3
npm ERR! code 128

npm ERR! Command failed: git fetch -a origin
npm ERR! remote: Repository not found.
npm ERR! fatal: repository 'https://github.com/haadcode/uport-persona-1.git/' not found
npm ERR!
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/richard/src/ipfs/distributions/dists/orbit/build/v3.2.0/orbit-3.2.0/npm-debug.log
npm ERR! Darwin 15.6.0
npm ERR! argv "/Users/richard/.nvm/versions/node/v6.4.0/bin/node" "/Users/richard/.nvm/versions/node/v6.4.0/bin/npm" "run" "dist-all"
npm ERR! node v6.4.0
npm ERR! npm  v3.10.3

npm ERR! missing script: dist-all
npm ERR!
[...]

PR #134, opened 2017-01-31
Add Orbit to dists

RichardLitt:
See https://github.com/ipfs/distributions/issues/133


lgierth:
Can you squash this into one commit once it's ready? Otherwise we end up with all these big files in history.

RichardLitt:
Yes yes, was planning to do that.

RichardLitt:
Ok. Almost done. The only issue now is that the files ought to be renamed to this format: 

`${id}_${version}_${osName}-${arch}.${ext}`

So, `orbit_master_darwin-amd64.tar.gz`. Does that work for you, @haadcode? If not, will refactor.

RichardLitt:
Conditional on #139, I believe this is good to go.

site/public/index.jade
```diff
@@ -2,7 +2,7 @@ include ./_mixins.jade
 
 include ./_header.jade
 
-- targetMap = {'386': '32-bit', 'amd64': '64-bit', 'arm': 'ARM'}
+- targetMap = {'386': '32-bit', 'amd64': '64-bit', 'x64': '64-bit', 'arm': 'ARM'}
```
dignifiedquire: why do we need to indroduce this additional alias?

scripts/mkdist.js
```diff
@@ -14,43 +14,40 @@ function mkdist (opts, cb) {
 
   var dist = opts.dist
   var ver = dist.version
-  if (ver.match(/^v/)) ver = ver.substring(1)
```
dignifiedquire: why are you removing this?

scripts/mkdist.js
```diff
@@ -14,43 +14,40 @@ function mkdist (opts, cb) {
...
 
-  dist.releaseLink = 'releases/' + dist.id + '/v' + ver
-  var releasePath = opts.path + '/' + dist.releaseLink
-  dist.platforms = []
+  dist.releaseLink = dist.id + '/' + ver
+  var releasePath = opts.path + '/releases/' + dist.releaseLink
+  dist.platforms = {}
```
dignifiedquire: why are you changng this to an object?

scripts/mkdist.js
RichardLitt: Because `master` as a version breaks it, and that's what we do with Orbit. I removed a line later down that had the `v` hardcoded, so it doesn't make a difference.

scripts/mkdist.js
RichardLitt: Because Orbit is the only app using mkdist.js, and it is stale in comparison to the JSON object built by `build-go.sh`. I had to change a few things to make it automatically add the archs we had; this was one of them.

site/public/index.jade
RichardLitt: Holdover from a previous time when we used x64. Removed.

scripts/mkdist.js
dignifiedquire: Not a big fan of adding master, releases should be tagged and have a version. Using something called `master` is completely unclear in what state it is

scripts/mkdist.js
dignifiedquire: 👍

scripts/mkdist.js
RichardLitt: For now, that is the build process for Orbit. When we are in a more stable process, we will move beyond `master`. 

@haadcode might have more to say on this matter (or he can link to somewhere where we've had this discussion before).

dignifiedquire:
@RichardLitt is there a built version of this on ipfs to look at?

RichardLitt:
https://ipfs.io/ipfs/Qma1W8UVavft5eBNE2ghynLpHD8dJg4eUwELKA81nZjCtk/ should work.

RichardLitt:
I also removed `ipget` from the build, as https://github.com/ipfs/distributions/issues/139 is open. We could also go with #140, too.

scripts/mkdist.js
dignifiedquire: `master` is just not a version, it should at least be a commit hash, if there is no proper version tagging.

RichardLitt: Yep. For now, though, that's how me and @haadcode agreed to do it. We can change this later.

dignifiedquire: @haadcode can you give some reasoning on that? How are people supposed to be able to differentiate builds, and updates on dist.ipfs.io if the version only says master?

haadcode: Sure. We don't have versioning nor a proper release process setup for orbit-electron, so we'll go with version "master" for now. This will change when we get the time to setup versioning in the Orbit's release process. Think of "master" as "latest" or "nightly".

haadcode:
> https://ipfs.io/ipfs/Qma1W8UVavft5eBNE2ghynLpHD8dJg4eUwELKA81nZjCtk/

^is not resolving anymore, can't download. @RichardLitt when you come online, can you make sure to pin it with pinbot? 

Also, it would be great if we could go through the whole UX flow with the built page: open it, click on all links, download Orbit for your platform and start it. All should succeed, nothing should be broken, easy :)

scripts/mkdist.js
VictorBjelkholm: @haadcode hm, even though there is no proper versioning, using the commit as the version would be the least you can do. I mean, in the end, master/latest/nightly is just a reference to a version (which still could be a commit). Simply providing "master" as a (mutable) version doesn't feel very user friendly and to be honest, be against what we want IPFS to stand for.

haadcode:
@RichardLitt found a commit-hash based solution here https://github.com/orbitdb/orbit-electron/pull/3. Ready to go from my side.

haadcode:
@RichardLitt made it even better, we'll release now with a version number, after all: https://github.com/orbitdb/orbit-electron/pull/4.

RichardLitt:
Ok. Seems to be working now. https://ipfs.io/ipfs/Qma4gdwsPjSYcsbP4Y8FVAWMQehb4bUnNwPVnv8zkUFiaM

dignifiedquire:
great to hear. @RichardLitt looking at the linked version it seems the header is somewhat broken, color is missing and it's not fixed anymore

RichardLitt:
I didn't change anything in the header, as far as I know. See the changed files. Any idea why those might have changed the header?

scripts/mkdist.js
whyrusleeping: This has to be a concrete version. It doesnt really matter if you "dont have a release process". Just pick a version and tag it.

The point here being that anybody should be able to pull down and build dists and end up with the same resultant binaries (ideally, though in reality builds arent always deterministic)

whyrusleeping:
the css for the site changed significantly (increased in size from 150kb to 200kb).

Which makes me wonder, why do we have 150kb of css. That seems rather excessive... >.>

RichardLitt:
> the css for the site changed significantly (increased in size from 150kb to 200kb).

It shouldn't have. This PR doesn't add any CSS. I'm not sure why it has changed.

scripts/mkdist.js
RichardLitt: We've implemented releases on GitHub. It works with this code.

dignifiedquire:
This might be because of changes in the dependencies, and they are not frozen currently

RichardLitt:
@dignifiedquire What would you suggest I do about the header?

dignifiedquire:
> @dignifiedquire What would you suggest I do about the header?

@whyrusleeping could you build this PR and see if you have the same issue?

dists/orbit/app-release
```diff
@@ -0,0 +1,92 @@
...
+  npm i
+  make
+  make build
+  make dist
+  mkdir -p bin/dist
+  cd bin/dist
+  touch orbit_0.1.0_darwin-amd64.tar.gz
+  touch orbit_0.1.0_linux-amd64.tar.gz
+
+  mkdir -p ../../../../../../../releases/orbit/$version
```
whyrusleeping: maybe put this path in an env var? makes errors in typing it out less likely

dists/orbit/app-release
```diff
@@ -0,0 +1,92 @@
...
+  cd $version
+  tmp=$(ls)
+  cd $tmp
+  npm i
+  make
+  make build
+  make dist
+  mkdir -p bin/dist
+  cd bin/dist
+  touch orbit_0.1.0_darwin-amd64.tar.gz
```
whyrusleeping: version is being hardcoded here, should we use the `$version` variable?

dists/orbit/app-release
RichardLitt: Shouldn't be any errors here, since it's already typed out.

dists/orbit/app-release
RichardLitt: Was an artefact of my testing the build. Removed.

whyrusleeping:
@dignifiedquire rebuilding the PR didnt reproduce the same issue until i deleted all my node_modules and reinstalled them. Now it looks the same as richards. 

I also had to deal with a bunch of other broken dependency shit. Can we please just set versions of things to a specific commit? I really don't like having to deal with mutable dependency bullshit that is all of the entire javascript ecosystem.

dists/orbit/app-release
```diff
@@ -0,0 +1,90 @@
...
+  tmp=$(ls)
+  cd $tmp
+  npm i
+  make
+  make build
+  make dist
+  mkdir -p bin/dist
+  cd bin/dist
+
+  mkdir -p ../../../../../../../releases/orbit/$version
```
whyrusleeping: Can we use a variable here for this?

dists/orbit/app-release
```diff
@@ -0,0 +1,90 @@
...
+  unzip "$version.zip" -d $version
+
+  echo "--> making $version"
+  cd $version
+  tmp=$(ls)
+  cd $tmp
+  npm i
+  make
+  make build
+  make dist
```
whyrusleeping: probably want to check if these fail and exit out with an error message:
```sh
if ! make build; then
  die "failed to run make build"
fi
```

dists/orbit/app-release
```diff
@@ -0,0 +1,90 @@
...
+stat "$build_path" > /dev/null 2>&1
+if [ "$?" -ne 0 ]; then
+  mkdir $build_path
+fi
+
+cwd=`pwd`
+
+release_version() {
+  # get version from input
+  version=$1
```
whyrusleeping: just to be careful, put `$1` in quotes

dists/orbit/app-release
```diff
@@ -0,0 +1,90 @@
...
+
+  echo "--> downloading $version"
+  wget $repo"/archive/"$version".zip" -P build
+  cd $build_path
+  unzip "$version.zip" -d $version
+
+  echo "--> making $version"
+  cd $version
+  tmp=$(ls)
+  cd $tmp
```
whyrusleeping: put quotes around all arguments to cd, any spaces in paths will break this

dists/orbit/app-release
VictorBjelkholm: Probably should have the hashbang line read `#!/bin/bash -e` and bash would automatically exit the whole script if anything fails.

Kubuxu:
There is program called `shellcheck`, it is great for linting and checking shell scripts.

diasdavid:
Is this still a WIP? @haadcode ??

haadcode:
I don't think is WIP anymore by anyone. Furthermore, the Electron version hasn't gotten any love since early this year and prolly doesn't work atm. I would postpone this until there's a new version and someone's actively maintaining it.

I'm fine closing this. @RichardLitt ?

PR #138, opened 2017-02-08
Remove ipfs-app

RichardLitt:
No one uses it, it breaks the build, it is already excluded, and I believe that it will not be resurrected in its current state.


VictorBjelkholm:
It's called "Station" nowadays, right? Not sure if we should remove it. Also, `dists_to_build=$(shell ls dists | grep -v ipfs-app)` indicates that it's already excluded from the build so nothing should be failing.

RichardLitt:
Can we agree on an upgrade path for including it in the build again? This seemed the easiest way to spark that conversation. 

I believe it is called Station. @dignifiedquire What is the status of Station?

VictorBjelkholm:
@RichardLitt I think it's a part of the "Stale Endeavors" that we should agree to do something about: https://github.com/ipfs/pm/issues/361

Removing it so the next time people won't find it on dist.ipfs.io might be a bit too fast for deprecating/handing something off.

RichardLitt:
> Removing it so the next time people won't find it on dist.ipfs.io might be a bit too fast for deprecating/handing something off.

It's not on dist.ipfs.io right now, as it is. The build process is also broken for it - I had to change a lot of things to make #134 work, to match the `build-go.sh` process used by all of the go packages.

dignifiedquire:
Yes, please remove it. What is there is nowhere close to usable as far as I can tell and doesn't even show up on dist.ipfs.io. If and when station moves forward we will have to set it up from scratch anyway.

Issue #139, opened 2017-02-08
ipget does not build


RichardLitt:
```
$ make ipget
echo "** ipget **"
** ipget **
cd dists/ipget && make
../../build-go.sh ipget github.com/ipfs/ipget versions
using /ipns/dist.ipfs.io as a template.
dirty output directory
will skip building already existing binaries
to perform a fresh build, please delete ../../releases/ipget
HEAD is now at f47d88c Add license to README.
building versions: v0.0.1
v0.2.0

v0.0.1 already exists, skipping...
Building version v0.2.0 binaries
==> checking out version v0.2.0 in /Users/richard/src/ipfs/distributions/dists/ipget/gopath/src/github.com/ipfs/ipget
HEAD is now at f47d88c Add license to README.
HEAD is now at f47d88c... Add license to README.
no matrix file for version v0.2.0 found, and no 'build_matrix' detected
make[1]: *** [dist] Error 1
make: *** [ipget] Error 2
```

RichardLitt:
@noffle Any ideas?

noffle:
It's missing a `build_matrix` file. You should be able to just `cp releases/gx/build_matrix releases/ipget`.

After this, there will still be a number of issues. ipget depends on `fallback-ipfs-shell`, which depends on some sub-packages of go-ipfs, which haven't been available under gx, so haven't been pinned. As a result, as the API of those sub-packages changes on go-ipfs master, the ipget build breaks. In order to get a stable build for ipget we'll need to start publishing sub-packages of go-ipfs to gx for `fallback-ipfs-shell` to use.

noffle:
For anyone interested in taking a stab at this, just clone github.com/noffle/ipfs-embedded-shell and do a `go build`: you'll see the breakages vs go-ipfs on master. You'll need to update the API calls to be consistent, and then create + publish those go-ipfs packages on gx, and add them as gx deps to ipfs-embedded-shell. You may need to repeat this process on `ipfs-fallback-shell` as well.

gammazero:
Is this still an issue?  The original repo [github.com/whyrusleeping/ipfs-embedded-shell](https://github.com/whyrusleeping/ipfs-embedded-shell) builds after updating dependencies:
```sh
git clone git@github.com:whyrusleeping/ipfs-embedded-shell.git
cd ipfs-embedded-shell
go mod edit -droprequire=github.com/ipfs/go-ipfs
go mod tidy
go build
```
[...]

PR #140, opened 2017-02-08
Filter out ipget v0.2.0

RichardLitt:
This currently does not build. By filtering, we allow the rest of the site to build without excluding ipget.


RichardLitt:
See #139.

whyrusleeping:
LGTM

PR #141, opened 2017-02-21
Less javascript is more

dignifiedquire:
- [x] Use yarn for dependency freezing
- [x] Migrate from harp (pretty much unmaintained) to hugo
- [x] Make sure everything still works


whyrusleeping:
@dignifiedquire does the output page look the same?

dignifiedquire:
yes, from what I can tell, but would like you to build and check for yourself just to be sure.

dignifiedquire:
Uploaded built version: https://ipfs.io/ipfs/QmcFcUNSxAaEt47fmsQgXvzXfbjH6nToiJLr8qpSfKd8Wm/

whyrusleeping:
@dignifiedquire maybe we should put something in the makefile to install hugo? Not a huge deal either way

dignifiedquire:
> @dignifiedquire maybe we should put something in the makefile to install hugo? Not a huge deal either way

I think that's OS dependent. We could check if it's in the path and print a message with the link to the instructions?

whyrusleeping:
@dignifiedquire that SGTM

whyrusleeping:
Built and it looks good to me :)

whyrusleeping:
@dignifiedquire merge when ready

dignifiedquire:
@whyrusleeping can you write this bash script, for checking for hugo? my bash foo is not getting any stronger today..

deps-check.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+function fail() {
+	printf $txtred%s$txtnon\\n "$@"
```
Kubuxu: This won't work as you expect. See the difference:

```
> f() {
printf '%s\n' "$@"
}
> f a b c
a
b
c
> f2() {
printf '%s\n' "$*"
}
> f2 a b c
a b c
```

Use "$*" instead of "$@".

dignifiedquire: fixed

PR #144, opened 2017-03-04
Add ipfs-cluster builds

hsanjuan:
This:

* Adds ipfs-cluster-ctl and ipfs-cluster-service
* Adds an optional "repo-name" file containing the name of the repository in case it does not match "distname" so that the links to source code and issues work for cluster.
* Rewrites two imports in dist.js without using "destructuring expressions" which make my js interpreter fail.
* Adds a few repo-owner files which were missing, thus fixing links to repositories in a few builds.

link: https://ipfs.io/ipfs/QmQu37NDZcRyWwpxvsRawCRmrAtoLfWMPg71k3gz9rjwU3


hsanjuan:
@whyrusleeping ok I think this is ready now....

dists/ipfs-cluster-service/description
```diff
@@ -0,0 +1,3 @@
+ipfs-cluster-service runs a full IPFS Cluster peer. These builds are a
+preview and are considered experimental. The last version of ipfs-cluster
```
whyrusleeping: s/last/latest/

scripts/dist.js
```diff
@@ -1,9 +1,10 @@
 'use strict'
 
 const fs = require('fs')
-const {join} = require('path')
+const join = require('path').join
```
whyrusleeping: did you need to change these for some reason?

build-go.sh
```diff
@@ -99,6 +99,7 @@ function doBuild() {
 	fi
 
 	# now package it all up
+        echo $dir/$binname
```
whyrusleeping: what is this echo for?

build-go.sh
```diff
@@ -138,7 +139,12 @@ function printInitialDistfile() {
 	local version=$2
 	test -e description || fail "no description file found"
 
-	printf "{\"id\":\"$distname\",\"version\":\"$version\",\"releaseLink\":\"/$distname/$version\"}" |
+	local reponame=$distname
```
whyrusleeping: quotes around $distname

build-go.sh
```diff
@@ -138,7 +139,12 @@ function printInitialDistfile() {
...
+	if [ -e repo-name ]; then
+		reponame=`cat repo-name`
```
whyrusleeping: I prefer using `$(thing)`

scripts/dist.js
whyrusleeping: nvm, read the top comment

build-go.sh
hsanjuan: good catch, leftover

hsanjuan:
Thanks @whyrusleeping. Addressed all comments.

PR #163, opened 2017-11-15
update readme

### LINKED ISSUE #157: Update README
The README is pretty out-of-date. Mainly, the background section. Unfortunately, it's *so* out of date that I can't easily figure out how it maps to the current system.

Also, The README mentions a versions file and the `Makefile` writes to this file. However, this file has *never* been checked into the repo so I'm not entirely sure what's happening here.

cc @whyrusleeping

Stebalien:
fixes #157

This also makes the Makefile no longer generate a `versions` file. According to
the old README, this was supposed to be checked into git but we never did
that...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ site: deps
 	npm run build
 
 publish: all_dists site
-	ipfs add -q -r releases | tail -n1 >>versions
+	ipfs add -q -r releases | tail -n1
 
 clean:
 	rm -rf releases
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ site: deps
 	npm run build
 
 publish: all_dists site
-	ipfs add -q -r releases | tail -n1
+	ipfs add -q -r releases | tail -n1 | tee -a versions
 
 clean:
 	rm -rf releases
--- a/README.md
+++ b/README.md
@@ -137,16 +137,14 @@ In the root of the repository, run:
 > make publish
 ```
 
-This will build (or download) anything that hasn't been built and build, compile the index in `releases`, and add releases to ipfs. Save the hash it spits out, that's the new hash for `dists.ipfs.io`.
+This will build (or download) anything that hasn't been built and build, compile the index in `releases`, and add releases to ipfs. Save the hash it spits out (we'll call it `<NEW_HASH>`), that's the new hash for `dists.ipfs.io`. We also append it to a file called `versions` in the repo root (*not* checked into git).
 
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -21,7 +21,7 @@ site: deps
 	npm run build
 
 publish: all_dists site
-	ipfs add -q -r releases | tail -n1 >>versions
+	ipfs add -q -r releases | tail -n1
```
hsanjuan: Actually, I found it useful to have a file which tracks which releases ive built.

wanna do `ipfs add -q -r releases | tee -a versions` ?

README.md
```diff
@@ -10,103 +10,150 @@
...
-- construct the missing `<dist>/<version>` directories
+If all looks well, pin the hash using pinbot (#ipfs-pinbot on Freenode, ask
+someone if you don't have permission to do so).
+
+Finally,
+
+1. Commit your changes and make a PR. Specifically, the changes to `dists/<dist>/versions` and `dists/<dist>/current`.
+2. File an issue on [ipfs/infrastructure](https://github.com/ipfs/infrastructure) with the hash you got from `make publish` and a link to the PR.
+
+If you have permission, you can just merge the PR, update the DNS, and then immediately, close the issue on ipfs/infrastructure. Ping someone on IRC.
```
hsanjuan: Add something about rememebering to pin the new hash with the pinbot, because I just forgot that step

Makefile
kevina: I agree with @hsanjuan

Stebalien: Sure (but I'm going to call it something else and add it to the gitignore).

README.md
Stebalien: Look up.

Makefile
Stebalien: Done.

README.md
hsanjuan: indeed

hsanjuan:
merge at your discretion

README.md
```diff
@@ -10,103 +10,150 @@
...
+
+This will build (or download) anything that hasn't been built and build, compile the index in `releases`, and add releases to ipfs. Save the hash it spits out, that's the new hash for `dists.ipfs.io`.
+
+Next, you should probably compare this with the current `dists.ipfs.io` to make sure nothing is amiss:
+
+```sh
+> ipfs object diff /ipns/dist.ipfs.io /ipfs/<NEW_HASH>
 ```
 
-Should:
```
kevina: I would also recommend you actually open the hash in your browser `http://127.0.0.1:8080/ipfs/HASH` and make sure that the latest version shows up and in general things look okay.

PR #164, opened 2017-11-17
update go-ipfs to v0.4.13

Stebalien:
Also, try to remove my home directory from builds as much as possible...


whyrusleeping:
two different trimpath flags?

Stebalien:
> two different trimpath flags?

Yes... even that's not quite sufficient. Technically, I should be cding into `/` or `/tmp` to make the builds fully deterministic.

Stebalien:
Note: one is passed to the assembler, the other to the `gc` (go compiler?) toolchain (linker?).

Issue #175, opened 2018-03-05
Add arm64 builds


Kubuxu:
The arm64 is quite popular right now so I think it warrants its own build.

magik6k:
Would fix https://github.com/ipfs/go-ipfs/issues/4583 too

renich:
Any progress on this one?

Stebalien:
@renich arm64 now builds but we have yet to start building it by default. Assuming we merge https://github.com/ipfs/distributions/pull/182, we should have support by the next release.

renich:
Cool! Thanks!

Issue #41, opened 2016-01-21
Clarify what ARM version we are supporting 


Dignifiedquire:
See https://en.wikipedia.org/wiki/List_of_ARM_microarchitectures for details

Dignifiedquire:
cc @whyrusleeping @jbenet

jbenet:
- does https://github.com/golang/go/wiki/GoArm imply we need to have builds for each? 
- https://golang.org/dl/ implies it's ARMv6...

not sure

PR #182, opened 2018-04-03
add arm64 support

Stebalien:
go-ipfs actually "supports" (it builds, at least) arm64. However, we don't currently build it.

Is this the correct way to add support?


lidel:
cc https://github.com/ipfs/distributions/issues/41, https://github.com/ipfs/distributions/issues/175

PR #227, opened 2019-03-20
Added executable bit on install.sh for tar.

Jorropo:
Tar is an unix friendly archive and allow to keep unix permission, so adding this is suposed install.sh to be executable right after `tar xf`.

### DIFF
--- a/build-go.sh
+++ b/build-go.sh
@@ -134,6 +134,9 @@ function bundleDist() {
 		return $?
 		;;
 	tar.gz)
+		if [ -f "$build_dir/install.sh" ]; then
+			chmod +x $build_dir/install.sh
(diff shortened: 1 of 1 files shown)


build-go.sh
```diff
@@ -134,6 +134,9 @@ function bundleDist() {
 		return $?
 		;;
 	tar.gz)
+		if [ -f "$build_dir/install.sh" ]; then
+			chmod +x $build_dir/install.sh
+		fi
```
Stebalien: We leave the executable bit as-set in the upstream repository. Where did you see this bit unset?

Jorropo:
executable bit is in tar, this was due to my unpacker.

PR #230, opened 2019-04-12
add go mod support

Stebalien:
Also, do a bit of cleanup. Please read the commits separately.


hsanjuan:
I did release cluster the other day with go-modules and without having to change anything in here... but other than that this seems ok.

build-go.sh
```diff
@@ -156,7 +175,7 @@ function printInitialDistfile() {
 	jq ".name = \"$distname\"" |
 	jq ".owner  = \"`cat repo-owner`\"" |
 	jq ".description = \"`cat description`\"" |
-	jq ".date = \"`date '+%B %d, %Y'`\"" |
+	jq ".date = \"`date '+%b %d, %y'`\"" |
```
magik6k: Why the lower case `%y`? I'd prefer `date '+%b %d, %Y'`

dists/fs-repo-migrations/Makefile
```diff
@@ -1,21 +1,22 @@
...
 all: dist
 
 dist: versions
-	../../build-go.sh fs-repo-migrations github.com/ipfs/fs-repo-migrations versions
+	../../build-go.sh ${distname} ${repo} ${package} versions
 	cp versions $(releases)/versions
 
 versions: filtered_versions
-	git ls-remote -t --refs $(repo) | egrep -o "refs/tags/v(.*)" | sed 's/refs\/tags\///' | sort > tag_versions
+	git ls-remote -t --refs https://$(repo) | egrep -o "refs/tags/v(.*)" | sed 's/refs\/tags\///' | sort > tag_versions
```
magik6k: We may also want to handle the `gx/vx.x.x` tags

magik6k: We also should export those common targets into a separate .mk file if possible

Stebalien:
> I did release cluster the other day with go-modules and without having to change anything in here... but other than that this seems ok.

You ended up building with `go get` (`strings ipfs-cluster-ctl | grep 'distributions/.*/gopath/'`).

build-go.sh
Stebalien: Thanks for catching that! I probably just lowercased the line.

dists/fs-repo-migrations/Makefile
Stebalien: We didn't rewrite the tags on go-ipfs, cluster, etc. (projects where we actually cut releases) so this shouldn't be an issue.

But yeah, we should have a common makefile.

Stebalien: Done.

hsanjuan:
> > I did release cluster the other day with go-modules and without having to change anything in here... but other than that this seems ok.
> 
> You ended up building with `go get` (`strings ipfs-cluster-ctl | grep 'distributions/.*/gopath/'`).

:( always assumed it called `make build`

PR #250, opened 2019-07-12
fix: don't mark pre-releases as "current"

Stebalien:


Stebalien:
That was sloppy, I had even commented out the `make` line. Fixed.

PR #272, opened 2019-09-25
rename darwin to macOS

### LINKED ISSUE #235: Dist page references "Darwin" instead of "MacOS" or "OS X"
// Bug reported by @tmcw
Screen Shot 2019-04-19 at 3 34 59 PM
Right now - the dist page lists a "darwin" binary instead of OS X or MacOS - which may confuse some users. I think this may be coming from the mkdist.js file, but I'm not sure where that's being called from. Looks like @dignifiedquire was the last person to touch this thoroughly and might have ideas? Happy to submit a PR if someone can give me the high-level sense of where all I need to make changes. =]

dignifiedquire:
This renames all visible mentions of "darwin" to "macOS". But the internal data structures and links still use darwin, to avoid breaking anything.

Closes #235

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,17 +13,17 @@
       }
     },
     "@babel/core": {
-      "version": "7.5.5",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.5.5.tgz",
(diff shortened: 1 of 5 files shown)


dignifiedquire:
cc @momack2 @jessicaschilling

jessicaschilling:
@dignifiedquire Thank you!

PR #284, opened 2020-03-05
Fix: set goos and goarch correctly

hsanjuan:
The shell linter thing must have broken this.


Stebalien:
@hsanjuan I've changed this to split using `read` so shellcheck is happy.

hsanjuan:
@Stebalien great, much better. I wonder if there is a shellcheck-happy of doing what I was doing...

PR #302, opened 2020-05-21
Includes fonts and amended .less accordingly

### LINKED ISSUE #62: Include fonts rather than linking from CDN
#### ETA April 14, 2020
Visual standardization was fixed via https://github.com/ipfs/distributions/pull/289, but we still need to include the fonts used in the site so it can be fully delivered over IPFS.

#### ETA April 2020
Standardize CSS and fonts against those used in www.ipfs.io and/or docs-beta.ipfs.io to a degree that brings them visually into line with existing branded materials, but stopping short of effort-heavy rebranding (we'll be tackling this as part of overall approach to ipfs.io in the future).

#### Original issue text
It should ideally be vendored

jessicaschilling:
- Includes Montserrat 200,300,400,500 roman and italic styles previously hosted on Google Fonts CDN
- Adds according @font-face declarations to base.less

Closes #62.


jessicaschilling:
@Stebalien, at your leisure. Thanks!

PR #307, opened 2020-06-10
update go ipfs v0.6.0 rc6

aschmahmann:
Also include support for older versions of jq.


aschmahmann:
closed in favor of #308

Issue #241, opened 2019-06-06
Add RSS feed of new releases to dist.ipfs.io


olizilla:
## This issue has a bounty!
Successfully closing this issue by making a PR of production-ready code can earn you not only the undying love of the IPFS community — it can net you a financial reward. [See the current list of bounty issues and their values here.](https://github.com/ipfs/devgrants/projects/1)

## How to claim this bounty
Want to work on this bounty project? Please leave a comment on this issue, and we'll be back in touch ASAP to discuss specifics.

## The need in brief
It'd be great to get an RSS feed for our releases. We have one on the GitHub releases page https://github.com/ipfs/go-ipfs/releases.atom but as dist.ipfs.io is our canonical reference, it would be better to offer a feed ourselves.

## The approach
Design:
1. Read in all `dist.json` files.
2. Sort by date.
3. Take the latest N (20?) releases.
4. Generate an atom feed, linked from the main dist page.

Language:
We can either do this in a shell script (run from the `site` target) or we can write a JavaScript program and manage it with NPM (probably easier to get an atom feed this way).

Stebalien:
Design:

1. Read in all `dist.json` files.
2. Sort by date.
3. Take the latest N (20?) releases.
4. Generate an atom feed, linked from the main dist page.

Language:

We can either do this in a shell script (run from the `site` target) or we can write a JavaScript program and manage it with NPM (probably easier to get an atom feed this way).

jay-dee7:
hey @olizilla, can i get to work on this issue over the weekend?

hsanjuan:
@jay-dee7 go ahead

jessicaschilling:
Hi @jay-dee7 - just checking in on this one? Thanks!

jay-dee7:
hey @jessicaschilling i've been flooded with work recently. I'll try my best to share an update this weekend

jay-dee7:
@jessicaschilling @olizilla just want to make sure that i'm thinking of the right away to approach this since i'm having a hard time understanding exactly what `read in file dist.json` means here.

I'm wondering i should collect all the releases from the github repos of ipfs' projects. then bundle them inside a json file with project name as the key, and then make the rss feed out of it (like an aggregated rss feeds from multiple repositories)

jessicaschilling:
@Stebalien -- I believe the approach came from your [note](https://github.com/ipfs/distributions/issues/241#issuecomment-517021669), are you able to advise @jay-dee7?

[...]

PR #312, opened 2020-08-17
Add RSS feed of new releases to dist.ipfs.io

neatonk:
See #241


jessicaschilling:
Hey @neatonk -- sorry for the delay on replying to this, I think I accidentally cleared a notification. Is it ready for review?

neatonk:
Yes, ready for review.

jessicaschilling:
Gateway link for review is here (built from my local of this PR):
https://gateway.ipfs.io/ipfs/QmQmvMzK1JFQyFS6YiKhTggr2wejLhLURH1AoSfmeRk1Jy

@Stebalien -- are you able to make a review? I've gone over it from the end-user perspective and it behaves as expected; just have one visual comment in my own review.

Thanks!

site/layouts/_default/single.html
```diff
@@ -12,8 +12,15 @@
   <body data-spy="scroll" data-target="#d-navbar" data-offset="40">
     {{ partial "header.html" }}
     <div class="container">
-      <div id="about" class="row d-about col-md-9 offset-md-3">
-        {{ partial "about.md" | markdownify }}
+      <div class="row">
+        <div id="about" class="d-about col-md-9 offset-md-3">
+          <div class="d-rss pl-3 float-right">
+            <a href="index.xml">
+              <img style="height: 1em" src="./img/icon-rss.svg">
```
Stebalien, line 19: Any reason not to specify the height in the CSS?

scripts/dist.js
```diff
@@ -31,6 +31,9 @@ function getVersion (type, done) {
 function writeData (type, version, done) {
   const dataPath = join(RELEASE_PATH, type, version, 'dist.json')
   const data = JSON.parse(fs.readFileSync(dataPath).toString())
+  if (!data.dateUTC && data.date) {
+    data.dateUTC = new Date(data.date).toUTCString()
```
Stebalien, line 35: I'd prefer to do this in the template rather than sneak it in here, but it's not a huge issue either way.

neatonk:
> This looks great overall! From examining https://gateway.ipfs.io/ipfs/QmQmvMzK1JFQyFS6YiKhTggr2wejLhLURH1AoSfmeRk1Jy (gateway-pinned instance from my local copy of this PR)

Thanks so much for building and pinning that.

> one small visual request: Possible to move the RSS icon/text onto its own line at narrower viewports so as to avoid awkward white space like the following?

Done. See https://github.com/ipfs/distributions/pull/312/commits/78227a4ea8e0263238d5463988302b4bce4baf5e.

Also here are some screenshots for review:

Screen Shot 2020-08-18 at 21 06 28
Screen Shot 2020-08-18 at 21 06 36
Screen Shot 2020-08-18 at 21 06 51
Screen Shot 2020-08-18 at 21 07 03
Screen Shot 2020-08-18 at 21 07 41
Screen Shot 2020-08-18 at 21 08 00

site/layouts/_default/single.html
neatonk: Yes, good question. Putting the height on the image tag ensures that the image will have a reasonable height, even if the CSS fails to load.

Stebalien: Ah, especially while loading. Got it.

scripts/dist.js
neatonk: Me too. I initially tried to use the [`dateFormat`](https://gohugo.io/functions/dateformat/) function provided by Hugo, but it cannot parse the date format used in the `data.json` files. 

Another option would be to change the date format used in the `data.json` files so they can be formatted as needed using the `dateFormat` function.

Stebalien: Ah... and we can't define a custom function. Eh, nevermind.

PR #324, opened 2020-12-09
remove darwin 386 support as it is not supported by Go 1.15+

aschmahmann:
Darwin 386 is not supported by Go 1.15+ (https://github.com/golang/go/issues/34749) so it should be removed from the build matrix of the binaries.


aschmahmann:
@olizilla @Stebalien giving you a heads up on this. Going to merge and push out a go-ipfs RC, but lmk if you have any issues with this.

Issue #326, opened 2020-12-11
Distribute repo migrations as set of binaries 


gammazero:
Currently, there is a single monolithic [binary](https://dist.ipfs.io/#fs-repo-migrations) responsible for all repo migrations, growing with each release that needs repo migration.  This will no longer be the case going forward.  Repo migrations will be available as a set of smaller binaries, one for each version migration (e.g. `ipfs-9-to-10`, `ipfs-10-to-11`). These will be downloaded and run individually by `go-ipfs`, `ipfs-update`, and `fs-repo-migration`.  To facilitate this, some changes to [dist.ipfs.io](https://dist.ipfs.io/) are needed.

### Directory structure changes
The plan is to store the store the migrations under a common root, `dist.ipfs.io/go-ipfs-repo-migrations/`:
```
go-ipfs-repo-migrations/
|-- ipfs-10-to-11/
|   |-- versions
|   `-- v1.8.0/
|       |-- ipfs-10-to-11_v1.8.0_darwin-amd64.tar.gz
|       |-- ipfs-10-to-11_v1.8.0_linux-amd64.tar.gz
|       `-- ipfs-10-to-11_v1.8.0_windows-amd64.zip
`-- ipfs-9-to-10/
    |-- versions
    `-- v1.8.0/
        |-- ipfs-9-to-10_v1.8.0_darwin-amd64.tar.gz
        |-- ipfs-9-to-10_v1.8.0_linux-amd64.tar.gz
        `-- ipfs-9-to-10_v1.8.0_windows-amd64.zip
```
The URL to retrieve a 10-to-11 migration for darwin is https://ipfs.io/ipns/dist.ipfs.io/go-ipfs-repo-migrations/ipfs-10-to-11/v1.8.0/ipfs-10-to-11_v1.8.0_linux-amd64.tar.gz

### Website changes
There should not be any visible changes to the [dist.ipfs.io](https://dist.ipfs.io/) web pages, since the current plan is to not show the users any links to individual migration binaries.  New versions of [`fs-repo-migrations`](https://dist.ipfs.io/#fs-repo-migrations) will continue to be available on the dist site, and will know how/where to download the necessary migration binary archives, unpack and install them.  There will not be a need for users to download individual migration binaries themselves.

Should it be necessary to download  a specific migration binary, a URL like the example above could be used.  If there ever is a need for users to do this, then a link from the main page to the top-level `go-ipfs-repo-migrations` could be added, but this does not appear to be needed for now.

[...]

PR #320, opened 2020-10-23
feat: only build versions that dont exist on dist

olizilla:
This PR reworks things to only build binaries that aren't already published on dist.ipfs.io. The releases dir now only needs to contain the files that are to change, and is treated as a patch to apply to the exist root CID for dist.ipfs.io.

I assert that this will work and be rad because content-addressing.

Also of note, this is a someewhat of a pre-requisite for #273 so we can start building dist in CI, and get binaries signed without requiring everyone to own apple hardware. On the default branch building dist takes about 25GB of disk and about 2hrs on a Quad-Core i7. It doesn't need to if we trust the hashes.

TODO:
- [x] only build what changed
- [x] fix website generation to not depend on a complete local releases dir
- [x] fix publishing to patch new release into existing dist.ipfs.io CID
- [x] Move the build step to CI, rather than expecting a maintainer to artisanally compile the bins on their machine

For subsequent PRs
- Add mac notarizing in CI as per #273 
- Push the diff to cluster
- Update the DNSLink from CI
- AUTOMATE ALL TEH THINGZ!!!1111!
  - Post go-ipfs binaries to the release from CI!?
  - watch for new semver tag on go-ipfs repo and re-build automatically!?

Fixes #260 

License: MIT
Signed-off-by: Oli Evans <oli@tableflip.io>

PR #327, opened 2021-01-12
Migrations

gammazero:
This PR creates individual distributions, for each IPFS fs repo migration.  This enables migration tools to download a minimal set of migrations needed to go from one version of IPFS to another.

The individual migrations will be available from the ipfs network, but are _not shown on the IPFS distributions web page_**.  Each migration has the same ipfs path structure as any other migration, so it can easily be found even in the absence of a link on the distributions web page.  For example, the `ipfs-10-to-11`  migration can be found at https://dist.ipfs.io/ipfs-10-to-11 and will have the same contents as other migrations.  NOTE: The migrations will not be published to IPFS until after approval/merge of this PR.

**This PR requires, and is built from, #320** 

**Addresses issue #326**

Additional modifications to the build scripts:
- Allow an optional sub-package to be specified when creating a new distribution
- Do not build web content if distribution contains `no-site` file
- Fix minor cleanup issues
- Fix CI issue with `jq` not being able to read files directly

** NOTE: There is a `no-site` file in each migration distribution that prevents building the web page content for that distribution.


jessicaschilling:
@Stebalien Are you able to review this, or pass along to someone else who can? Thank you!

scripts/patch.js
```diff
@@ -0,0 +1,105 @@
...
+const all = require('it-all')
+const chalk = require('chalk')
+const IpfsHttpClient = require('ipfs-http-client')
+const { globSource } = IpfsHttpClient
+require('make-promises-safe') // exit on error
+
+const ipfs = IpfsHttpClient()
+const pathTo = (file) => path.join(__dirname, '..', file)
+
+const DIST_DOMAIN = 'dist.ipfs.io'
```
Stebalien, line 29: nit: I'd consider using a random string `dist-.....` until we're done so we don't accidentally clobber something (and so we know the process "completed").

Stebalien (approved):
Looks reasonable to me. I didn't check every change in detail, but if it works, it works. And if something goes wrong, a manual "diff" should catch that.

build-go.sh
```diff
@@ -315,15 +315,26 @@ function startGoBuilds() {
...
+	if [ ! -e "$versions" ]; then
+		fail "versions file $versions does not exist"
+	fi
+
+	if ! mkdir -p "$releases"; then
+		fail "failed to create releases directory"
+	fi
+
 	if [ -z "$existing" ]; then
 		existing="/ipns/dist.ipfs.io"
```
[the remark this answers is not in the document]
gammazero: Done; Done

build-go.sh
```diff
@@ -315,15 +315,26 @@ function startGoBuilds() {
 	versions="$4"
 	existing="$5"
 
+	if [ ! -e "$versions" ]; then
+		fail "versions file $versions does not exist"
+	fi
+
+	if ! mkdir -p "$releases"; then
```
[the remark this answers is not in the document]
gammazero: I think a temp dir is preferable, but that would need to be communicated to other scripts that build the site information.  For that change, I would feel more comfortable handling that in a separate PR.

scripts/patch.js
gammazero: I think it is necessary to use the real domain, since that is resolved to the current distribution, and then used to calculate what needs to be patched.

jessicaschilling:
@olizilla Are you wanting to wait to merge this until #320?

scripts/patch.js
Stebalien: Ah, sorry, I put the comment in the wrong location. I meant use a random string for MFS_DIR.

Stebalien: In case I, e.g., ran a partial build and failed somewhere along the way.

olizilla:
This PR builds on top of #320, so we should merge #320 first. More generally for merge decisions on this repo, I defer to Adin as he has to use this to release go-ipfs and Steven as the most prolific contributor.

scripts/patch.js
gammazero: Appended ISO data string to `dist.ipfs.io`, to be unique plus indicate when dist was published.
```
> ipfs files ls
dist.ipfs.io_2021-02-01T21:50:30.407Z
```

scripts/patch.js
```diff
@@ -26,8 +26,9 @@ require('make-promises-safe') // exit on error
 const ipfs = IpfsHttpClient()
 const pathTo = (file) => path.join(__dirname, '..', file)
 
+const event = new Date()
 const DIST_DOMAIN = 'dist.ipfs.io'
-const MFS_DIR = `/${DIST_DOMAIN}`
+const MFS_DIR = `/${DIST_DOMAIN}` + '_' + event.toISOString()
```
aschmahmann, line 31: This means that people who run this will end up with one MFS reference per-build instead of one per dist website. Given that my understanding is we're pinning all the files we add anyway the only problem that could come up with this is just it polluting your MFS root space.

nit: WDYT about having some nesting to keep `ipfs files ls /` cleaner? An example might be `/${DIST_DOMAIN}/${DIST_DOMAIN}` + '_' + event.toISOString()`

Makefile
```diff
@@ -24,6 +24,7 @@ publish: deps all_dists site
 
 clean:
 	rm -rf releases
+	chmod -f -R u+w dists/*/gopath || true
```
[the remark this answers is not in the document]
Stebalien: Go likes making some things read-only, and that can cause problems for `rm -rf` on some platforms.

build-go.sh
```diff
@@ -57,9 +60,9 @@ function buildDistInfo() {
 	local dir="$2"
 
 	# print json output
-	if ! jq -e ".platforms[\"$goos\"]" dist.json > /dev/null; then
+	if ! jq -e ".platforms[\"$goos\"]" < dist.json > /dev/null; then
```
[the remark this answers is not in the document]
gammazero: Yes, that is what that was for and was all that was needed.

build-go.sh
```diff
@@ -324,11 +327,11 @@ function startGoBuilds() {
 	fi
 
 	if [ -z "$existing" ]; then
-		existing="/ipns/dist.ipfs.io"
+		existing="$DIST_PATH"
 	fi
 
 	echo "comparing $versions with $existing/$distname/versions"
-	newVersions=$(comm -13 <(ipfs cat "$existing/$distname/versions") "$versions")
+	newVersions=$(comm --nocheck-order -13 <(ipfs cat "$existing/$distname/versions") "$versions")
```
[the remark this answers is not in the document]
gammazero: This fixes an issue when trying to build a distribution that needs semver sorting.  Without the `--nocheck-order` flag, sorting breaks the diff check because the following:
```
v0.1.0
v0.2.0
v0.10.0
```
gets incorrectly reordered as:
```
v0.1.0
v0.10.0
v0.2.0
```
Without this fix, you cannot build https://dist.ipfs.io/ipfs-cluster-service

dist.sh
```diff
@@ -37,15 +38,12 @@ case $1 in
 		mkdir -p "dists/$name"
 
 		cp templates/build_matrix "dists/$name/"
-		sed "s/ABCGHREPOXYZ/$(sedEscapeArg "$repo")/g" templates/Makefile | sed "s/ABCDISTNAMEXYZ/$name/g" > "dists/$name/Makefile"
+		sed "s/github.com\/foo\/bar/$(sedEscapeArg "$repo")/g" templates/Makefile | sed "s/cmd\/bar/$subpkg/g" > "dists/$name/Makefile"
```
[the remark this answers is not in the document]
gammazero: Yes.  I was a bug leftover from when the string `ABCDISTNAMEXYZ` appeared in the Makefile template.

dist.sh
```diff
@@ -37,15 +38,12 @@ case $1 in
...
 		echo "$description" > "dists/$name/description"
 		echo "$latest_tag" > "dists/$name/current"
 		echo "$latest_tag" > "dists/$name/versions"
-		echo "" > "dists/$name/filtered_versions"
```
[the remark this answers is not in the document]
gammazero: Yes, because that file is never used.

dist.sh
```diff
@@ -37,15 +38,12 @@ case $1 in
...
 
-		echo "distribution $name created successfully! starting build..."
-
-		make "$name"
+		echo "distribution $name created successfully! To start build: make $name"
```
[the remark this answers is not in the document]
gammazero: Yes, and there may be some cases where you want to examine things before running make.

templates/build_matrix
```diff
@@ -1,4 +1,3 @@
-darwin 386
```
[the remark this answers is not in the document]
gammazero: Yes

scripts/patch.js
gammazero: I actually want to pollute MFS root.  The is so I see that I have multiple dists and remember to remove those that I no longer need.  This is not for the purpose of archiving dists, but to avoid stepping on a previous one.  Once I am happy with a dist, I expect to remove previous ones.

@Stebalien suggested in the previous review, that I use a random string here. I decided that date was more useful to determine which was the most recent.

Stebalien: Honestly, keeping old ones should be _fine_ as it shouldn't actually use any more data. It also makes it easier to delete them (delete the root).

But either way. I don't have strong opinions.

aschmahmann: this is just a local thing for the builder we can always change later if it becomes a problem with no significant consequences, so we can just leave it as is and figure it out.

aschmahmann:
This PR looks good to go, going to merge since it has a few bug fixes in it anyway and if we find anything about the `no-site` binaries we want to change that can be a separate PR.

scripts/patch.js
olizilla: this is fine and good. I am compelled to link to the fledgling "cohosting" spec which has an opinion on MFS dir names for snapshots of dnslink'd websites https://github.com/ipfs-shipyard/cohosting/blob/master/SPEC.md#path-conventions. It's only mildly relevant here as the goal of this script is not co-hosting the dist site, but is worth knowing about.

PR #331, opened 2021-01-29
release: fs-repo-migrations v1.7.1

aschmahmann:
CID: QmYRLRDKobvg1AXTGeK5Xk6ntWTsjGiHbyNKhWfz7koGpa

Note: This PR also adds the ipfs-cluster v0.13.0-rc1 set of binaries (cc @hsanjuan)


hsanjuan:
@aschmahmann those should have been deleted (they are not in the `versions` file). I mean, I think I had deleted them in my build. The problem is, unless you have cleaned up the build folder (manually) before running `make`, if you had any old binaries there (like 0.13.0-rc1), they will be included in the final result even if not in `versions`.

Anyways, we'll clean them up next time.

aschmahmann:
@hsanjuan the problem is that the script looks at all the tags and tries to build them, so since there was a tag it tried to build it unless you did something weird. Any problems just leaving it, otherwise we can try and remove it.

PR #334, opened 2021-03-23
Rename migrations from ipfs-x-to-y to fs-repo-x-to-y

gammazero:
Fix build-go.sh to build using go modules, instead of GOPATH.


dists/fs-repo-0-to-1/current
```diff
@@ -0,0 +1 @@
+v2.0.0
```
aschmahmann, line 1: Why are these all v2?

build-go.sh
```diff
@@ -6,6 +6,9 @@ set -eo pipefail
 GOPATH="$(go env GOPATH)"
 export GOPATH
 
+# Always use go modules
+export GO111MODULE=on
+
```
aschmahmann, line 11: Does the build process now rely on Go modules in order to function here?

dists/fs-repo-0-to-1/current
gammazero: Previously, these did not have their own versions, so they inherited the version of the fs-repo-migrations repo.  Since we are going to version the fs-repo-migration to 2.0.0 for next release, I thought that we would want the individual migrations versioned likewise.  We can choose any versioning we want though, nothing depends on knowing their version.

build-go.sh
gammazero: No, but we should not have any more go repos that do not use modules.  Also, `GOPATH` (and therefore also `GO111MODULE`) will not be supported in go1.17.  So, I would like the build to fail if anything still needs `GOPATH` or does not have module support.

PR #340, opened 2021-04-15
remove perl dependency

Stebalien:
On linux, sha512sum is a part of the "coreutils" package. shasum is a part of perl.


gammazero (approved):
Nice!

Issue #350, opened 2021-05-28
Build fails on macos due to unsupported `comm` flag `--nocheck-order`


olizilla:
The version of `comm` that ships with macos does not support the `--nocheck-order` flag which got added in https://github.com/ipfs/distributions/commit/19448fd1a5098fde6ce4f7c24d74a30884bba7e4

so macos users get

```console
** fs-repo-0-to-1 **
/Library/Developer/CommandLineTools/usr/bin/make -C dists/fs-repo-0-to-1
../../build-go.sh "fs-repo-0-to-1" "github.com/ipfs/fs-repo-migrations" "fs-repo-0-to-1" versions
comparing versions with /ipfs/QmUeJDaHiaETrpyqaydtp5Fgbkayz6qxkJ7n2Ddyi4pbpq/fs-repo-0-to-1/versions
comm: illegal option -- -
usage: comm [-123i] file1 file2
make[1]: *** [dist] Error 1
make: *** [fs-repo-0-to-1] Error 2
```

olizilla:
This is so dull. This should probably be dockerized.

aschmahmann:
@olizilla note that building in docker is almost certainly a good idea given #348 as well as #322

lidel:
Closed by #352 and https://github.com/ipfs/distributions/pull/367 
TLDR: run `./dockerized make publish` or ./dockerized ./dist.sh`

Issue #343, opened 2021-05-03
TextDecoder is not defined


hsanjuan:
This happens now (also with a checkout from scratch):

```
** Building site **
npm run build

> dist.ipfs.io@1.0.0 build /home/hector/ipn/distributions
> npm run dist && npm run build:assets && npm run build:hugo


> dist.ipfs.io@1.0.0 dist /home/hector/ipn/distributions
> node scripts/dist.js

/home/hector/ipn/distributions/node_modules/cids/node_modules/multibase/src/util.js:3
const textDecoder = new TextDecoder()
                    ^

ReferenceError: TextDecoder is not defined
    at Object.<anonymous> (/home/hector/ipn/distributions/node_modules/cids/node_modules/multibase/src/util.js:3:21)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/home/hector/ipn/distributions/node_modules/cids/node_modules/multibase/src/base.js:3:24)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! dist.ipfs.io@1.0.0 dist: `node scripts/dist.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the dist.ipfs.io@1.0.0 dist script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/hector/.npm/_logs/2021-05-03T16_31_24_090Z-debug.log
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! dist.ipfs.io@1.0.0 build: `npm run dist && npm run build:assets && npm run build:hugo`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the dist.ipfs.io@1.0.0 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
[...]

PR #352, opened 2021-06-02
Dockerize

### LINKED ISSUE #351: Build binaries in Docker containers
Instead of building and packaging the binaries on the local machine it could be very useful to do these in docker containers.

This would help with reproducible binary builds https://github.com/ipfs/go-ipfs/issues/7848 as well as making life easier for deployers when dealing with issues like #348 #343  #350  and even make it easier to do #322.

thattommyhall:
Should fix #351, work-around for #350 if they use Docker instead, addresses how to agree versions from #343

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
 nodejs 16.2.0
-golang 1.16.4
+golang latest:1.16
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
 nodejs 16.2.0
-golang latest:1.16
+golang 1.16.5
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,0 +1,2 @@
+nodejs 16.2.0
+golang 1.16.4
--- a/dockerized
+++ b/dockerized
@@ -5,7 +5,7 @@ set -euxo pipefail
 docker pull ubuntu:20.04
 
 # CACHEBUST means this will apply the updates once a day
-docker build . -t distributions --build-arg CACHEBUST=`date --iso-8601=date` 
+docker build . -t distributions --build-arg CACHEBUST=`date --iso-8601=date` --build-arg USER_UID=$(id -u "$USER")
 
 # We use host networking as the build process assumes a fairly long-lived ipfs
 # node has the CIDs (we give them to the collab cluster to pin) 
--- a/dockerized
+++ b/dockerized
@@ -1,0 +1,12 @@
+#!/bin/bash
+set -euxo pipefail
+
+# We always want the latest image
+docker pull ubuntu:20.04
+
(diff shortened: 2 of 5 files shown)


dockerized
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -euxo pipefail
+
+# We always want the latest image
+docker pull ubuntu:20.04
+
+# CACHEBUST means this will apply the updates once a day
+docker build . -t distributions --build-arg CACHEBUST=`date --iso-8601=date`
```
gammazero, line 8:
```suggestion
docker build . -t distributions --build-arg CACHEBUST=`date --iso-8601=date` --build-arg USER_UID=$(id -u "$USER")
```
The docker container will not have permission to write to the mapped volumes since it is running as the container's `asdf` user, which has the uig:gid 1000:1000.  This means that the user needs to make their `distributions` directory (mapped into the container as `/build`) accessible to that uid. **Or, better, create the `asdf` user with the same uid that the user invoking this script has.** This passes the user's uid into the Dockerfile as a build arg.

gammazero (changes requested):
In order for this to run correctly on most users' systems, a few small changes are needed.

gammazero (approved):
LGTM

.tool-versions
```diff
@@ -0,0 +1,2 @@
+nodejs 16.2.0
+golang 1.16.4
```
gammazero, line 2:
```suggestion
golang latest:1.16
```
This will use the latest version of golang 1.16.x.  I tested this and saw it picked up 1.16.5

dockerized
```diff
@@ -0,0 +1,12 @@
...
+
+# We always want the latest image
+docker pull ubuntu:20.04
+
+# CACHEBUST means this will apply the updates once a day
+docker build . -t distributions --build-arg CACHEBUST=`date --iso-8601=date` --build-arg USER_UID=$(id -u "$USER")
+
+# We use host networking as the build process assumes a fairly long-lived ipfs
+# node has the CIDs (we give them to the collab cluster to pin) 
+docker run -it --network host -v `pwd`:/build distributions "$@"
```
gammazero, line 12:
```suggestion
docker run --rm -it --network host -v `pwd`:/build distributions "$@"
```
There should not be any need to keep the exited container around after the build, so clean it up after done with docker container.

gammazero:
@thattommyhall Added a couple of other suggestions post-review.

.tool-versions
thattommyhall: I might have been hasty agreeing to this, if this file does not change then the docker image wont be updated as the cache wont be invalidated and I dont know the behavior there on run (will it prompt to update?)

You folks have to live with it so whatever works best for you

Issue #330, opened 2021-01-29
RSS feed times are effected by timezone of builder


aschmahmann:
Despite all the times being published as UTC two different builds result in different times. For example,

```
<       <pubDate>Wed, 06 Mar 2019 23:00:00 GMT</pubDate>
---
>       <pubDate>Thu, 07 Mar 2019 05:00:00 GMT</pubDate>
47c50
<       <pubDate>Sun, 07 Oct 2018 22:00:00 GMT</pubDate>
---
>       <pubDate>Mon, 08 Oct 2018 04:00:00 GMT</pubDate>
55c58
<       <pubDate>Wed, 13 Jan 2021 23:00:00 GMT</pubDate>
---
>       <pubDate>Thu, 14 Jan 2021 05:00:00 GMT</pubDate>
63c66
<       <pubDate>Wed, 13 Jan 2021 23:00:00 GMT</pubDate>
---
>       <pubDate>Thu, 14 Jan 2021 05:00:00 GMT</pubDate>
71c74
<       <pubDate>Wed, 13 Jan 2021 23:00:00 GMT</pubDate>
---
>       <pubDate>Thu, 14 Jan 2021 05:00:00 GMT</pubDate>
79c82
<       <pubDate>Wed, 30 Sep 2020 22:00:00 GMT</pubDate>
---
>       <pubDate>Thu, 01 Oct 2020 04:00:00 GMT</pubDate>
87c90
<       <pubDate>Sun, 10 Dec 2017 23:00:00 GMT</pubDate>
---
>       <pubDate>Mon, 11 Dec 2017 05:00:00 GMT</pubDate>
95c98
<       <pubDate>Mon, 10 Oct 2016 22:00:00 GMT</pubDate>
---
>       <pubDate>Tue, 11 Oct 2016 04:00:00 GMT</pubDate>
103c106
<       <pubDate>Wed, 22 Apr 2020 22:00:00 GMT</pubDate>
---
>       <pubDate>Thu, 23 Apr 2020 04:00:00 GMT</pubDate>
111c114
<       <pubDate>Mon, 11 May 2020 22:00:00 GMT</pubDate>
---
>       <pubDate>Tue, 12 May 2020 04:00:00 GMT</pubDate>
```

Which is conveniently the timezone difference between me and the previous publisher.

jessicaschilling:
@aschmahmann Do you have the bandwidth to PR a fix for this?

aschmahmann:
[...]

PR #353, opened 2021-06-23
Set container timezone to UTC

gammazero:


marten-seemann (commented):
Why LA, why not UTC?

gammazero:
@marten-seemann PL has headquarters in pacific time, but really that can be anything.  If we are always going to use UTC, which should be the container default, do we really need this PR?

marten-seemann:
> If we are always going to use UTC, which should be the container default, do we really need this PR?

I'm not sure which is issue is fixed by this PR, you tell me ;)

While it may be true that PL is headquartered in Pacific Time, it very much understands itself as an international (interplanetary?) company. The default timezone, when no other is applicable, is UTC, which is why I suggested it.

aschmahmann:
> If we are always going to use UTC, which should be the container default, do we really need this PR?

We should run some tests to see if/how the new docker containers fix our problems. We've only done one release using the containers and it's on my machine. If you try doing a local build of the website and diff with dist.ipfs.io and you don't see lots of changes in the XML file of the RSS feed #330 then we're already good to go. Otherwise, something's not going quite right.

lidel (approved):
Given that we merged #369 and #367 and we default to UTC all over the place now, 
I rebased this PR to switch to use  UTC in Docker as well.

Issue #348, opened 2021-05-14
Dealing with glibc linking and CGO


aschmahmann:
glibc v2.32 seems to have breaking changes compared to prior versions of glibc. This means that if the binaries are built with glibc >= 2.32 they will not be compatible with prior versions. For example, building on stock Ubuntu 20.10 will result in the build not running on Ubuntu 20.04.

The main options seem to be:
1. Disable CGO
  - not an option for all binaries (e.g. go-ipfs requires them for plugins)
 2. Statically link glibc (and all associated C libraries)
   - IIUC just linking glibc itself can cause a number of problems due to their still being some reliance on the system's glibc
   - static linking all the C dependencies won't work for plugins
3.  Build in an environment where older glibc is supported (e.g. glibc = 2.31 as in Ubuntu 20.04)

Note: while go-ipfs could drop support for external plugins we might still run into issues if we want to include other C libraries such as databases, WASM VMs, etc.

For now the easiest thing to do is to is option 3 by spinning up a VM. However, having a standardized build environment seems like a reasonable thing anyway perhaps we should go back (I heard it may have been this way a number of years ago) to building inside of docker containers.

Update: We should also run some script in this repo like https://gist.github.com/fasterthanlime/17e002a8f5e0f189861c to verify that the glibc versions used are below whatever our support threshold is.

cc @mvdan @hsanjuan @gammazero

mvdan:
Docker containers is what I would do; one could start with images like golang:1.16.2-buster, since that's Debian stable and includes glibc 2.28. A container for Go builds is unnecessary for pure Go builds, but if we want cgo, one needs some way to pin the versions of C libraries.

>  IIUC just linking glibc itself can cause a number of problems due to their still being some reliance on the system's glibc

For static linking of C libraries, I'd consider musl instead, which is actually designed with static linking in mind. It also works well with Go, since it's libc-compatible (but not glibc-compatible).

> static linking all the C dependencies won't work for plugins

I'd be interested to hear more - I'd imagine this should work.

aschmahmann:
> I'd be interested to hear more - I'd imagine this should work.

How? If the plugins are in-tree then CGO isn't required, and if the plugins are out-of-tree then you don't have them available at compile time to perform the static linking.

mvdan:
Perhaps I misunderstood. By static linking I mean to cgo libraries, like libc. Go plugins are never statically linked, as that would defeat their purpose - they'd just become normal Go packages.

[...]

PR #362, opened 2021-07-13
feat: macos signing and notarization

lidel:
This PR is a step towards closing https://github.com/ipfs/go-ipfs/issues/8240 globally (outside of Brave)

# This PR is WIP, do not merge 

## TODO

- [x] test setup to re-build go-ipfs 0.9.0 on CI on every `push` – easy test on PR without merging to master (iiuc `workflow_dispatch` work only from main branch)
- [x] `workflow_dispath` that allows building new versions of existing dists
- [x] `add-version` working on CI 
- [ ] persist final CID by pinning to cluster
  - needs https://github.com/ipfs-shipyard/ipfs-github-action/pull/7
  - needs  https://github.com/protocol/bifrost-infra/issues/1326
- [x] run `./dockerized` version to match binaries built for docker images in go-ipfs repo
- [ ] run ABI test script from  https://github.com/ipfs/distributions/issues/348
  - TODO: extract to separate job that runs in parallel to macos notarization + also runs on the main CI workflow 
- [x] make sure `dist.json` and `.cid` and `.sha512` gets updated after signing/notarization
- [x] (could be separate PR) introduce framework for signing (windows / macos)
  - Q: do we need separate appid?  any risks if we reuse ones from ipfs-desktop?
    -  A: better safe than sorry: solved it in safer way – we use unique `bundle_id` per dist package
- [ ] open problem: `add-version` mutates git repo – figure alternative approach
  - we could simplify workflow by keeping `add-version` local and including signing as part of `main.yml`
     - this way, `./dists.sh add-version` is run locally, then developer opens a PR which includes change to `dists/*/versions` and `main.yml` is executed against that
     - new version won't be present at `/ipns/dist.ipfs.io` and will be built during `make all_dists` which is a prerequisite for `make publish`
     - we should be able to make it work by 
       - having an explicit _build_ and _sign_ stages before existing _publish_ one
       - the challenge is to make singing smart – detect which new packages are present in `./releases`  and run for each 
       - test setup:  env variable that replaces `/ipns/dist.ipfs.io` with artificial site that is missing binaries – this way `make all_dists` will create "missing" binaries in `./releases` and we can reuse signing logic from `add-release.yml` 
- [ ] remove test setup on `push`
- [ ] update README around the rube goldberg machine


lidel:
Superseded by https://github.com/ipfs/distributions/pull/367

PR #366, opened 2021-07-20
feat: Add deep diffs to PRs

guseggert:
- add scripts/diff.js which finds the changed objects and then diffs
them all
- on a pull request, posts the diff as a comment on the PR


scripts/diff.js
```diff
@@ -0,0 +1,66 @@
...
+
+const DIST_ROOT = '/ipns/dist.ipfs.io'
+const CHANGE_TYPE_ADD = 0
+const CHANGE_TYPE_REMOVE = 1
+const CHANGE_TYPE_MOD = 2
+
+// Set of files to skip printing diffs.
+// These typically contain minified code that produce unreadable diffs.
+// Fail open so we don't accidentally miss something.
+const skipDiff = new Set(['site.js', 'site.css'])
```
aschmahmann, line 18: Yep, diffs of minified content are annoying. However, we probably want to at least know that these files are changed so that we can check on them when they're expected (e.g. bumping one of our JS deps) and verify they're not happening when unexpected.

aschmahmann (commented):
I'm a little concerned that this diff is going to show up in every PR we make and cause noise. The `lastBuildDate` is always going to be updated and is something we'd just have to ignore, but maybe is livable. Is it a huge pain to ignore that one part of the diff?

If #330 continues to be a problem and we show a huge diff on every PR that's pretty sure to start bothering people. I can give it a try along with the go-ipfs v0.9.1 release though and see how that goes (hopefully building in docker resolves #330 and makes this not a problem).

Worst comes to worst we can make this action triggerable by command instead of happening on every PR.

scripts/diff.js
guseggert: Good point, I'll add a list at the beginning of files that changed but whose diffs are not shown.

guseggert:
> Is it a huge pain to ignore that one part of the diff?

No that's not hard, I'll add that.

guseggert:
You can see above that the dates are now skipped and it just leaves a "nothing changed" message.

Here's from a sample PR that shows what it now looks like when there are substantive changes:

## Diff of Changes

The following files changed but will not show diffs, because the diffs would be unreadable:
* site.css
* site.js

```diff
diff -u --recursive --ignore-matching-lines '^\s*<pubDate>.*</pubDate>\s*' --ignore-matching-lines '^\s*<lastBuildDate>.*</lastBuildDate>\s*' a/404.html b/404.html
--- a/404.html 2021-07-21 16:16:07.022398201 +0000
+++ b/404.html 2021-07-21 16:16:07.158405984 +0000
@@ -1,2 +1,3 @@
 <h1>404</h1>
 <h3>Whoops. Looks like what you're looking for can't be found.</h3>
+this is some new text
diff -u --recursive --ignore-matching-lines '^\s*<pubDate>.*</pubDate>\s*' --ignore-matching-lines '^\s*<lastBuildDate>.*</lastBuildDate>\s*' a/index.html b/index.html
--- a/index.html 2021-07-21 16:16:07.314414911 +0000
+++ b/index.html 2021-07-21 16:16:07.426421320 +0000
@@ -30,6 +30,7 @@
 Available as RSS</a>
 </div>
 <h4 id="welcome-to-ipfs-distributions">Welcome to IPFS Distributions</h4>
+<p>This is some new text.</p>
 <p>This is the downloads website for all the official software distributions of the
 <a href="https://ipfs.io">IPFS Project</a>. You can find all the apps, binaries, and
 packages here. Every distribution has a section on this page with &hellip;</p>

```

scripts/diff.js
```diff
@@ -0,0 +1,116 @@
...
+      default:
+        throw new Error(`unexpected diff type ${f.Type}`)
+    }
+  }
+
+  const diffResult = await spawnAsync('diff', [
+    '-u',
+    '--recursive',
+    // these change on every build, so they just add noise, which is why we ignore them
+    '--ignore-matching-lines', '^\\s*<pubDate>.*</pubDate>\\s*',
```
aschmahmann, line 86: I think we don't need to ignore this one since any noise here is probably due to a bug (e.g. #330)

thattommyhall:
lastBuildDate thing is potentially better fixed by #370

scripts/diff.js
```diff
@@ -0,0 +1,122 @@
...
+'use strict'
+
+const spawn = require('child_process').spawn
+const fs = require('fs/promises')
+const path = require('path')
+const util = require('util')
+const exec = util.promisify(require('child_process').exec)
+require('make-promises-safe') // exit on error
+
+const DIST_ROOT = '/ipns/dist.ipfs.io'
```
lidel, line 11: #367  will need this for consistency during some custom runs:

```suggestion
const DIST_ROOT = process.env.DIST_ROOT || '/ipns/dist.ipfs.io'
```

scripts/diff.js
```diff
@@ -0,0 +1,122 @@
...
+        throw new Error(`unexpected diff type ${f.Type}`)
+    }
+  }
+
+  const diffResult = await spawnAsync('diff', [
+    '--new-file',
+    '-u',
+    '--recursive',
+    // this changes on every build, so just adds noise, which is why we ignore it
+    '--ignore-matching-lines', '^\\s*<lastBuildDate>.*</lastBuildDate>\\s*',
```
lidel, line 89: Won't be needed after #370 is merged
```suggestion
```

lidel (changes requested):
Thanks, this is very useful PR in the context of us moving to releases driven by PRs in this repo :+1: 

We can't merge this yet, as other things need to land first:

- [ ] merge `lastBuildDate` and `pubDate` fixes from #370 – this will remove the need for special handling of  here
- [ ] merge updated CI with signing from #367  – `main.yml` was rewritten, so we will just copy dif job into the new one
- [ ] merge nightly builds from #369  – PR includes some fixes around UTC dates which would not hurt
- [ ] rebase this PR on top of above changes

I will be doing this dance next week, and can take care of refactoring (I already have all the context).

PR #367, opened 2021-07-21
feat: macos signing and notarization 

lidel:
> This is a streamlined  refactor of #362 on road to fix issues described in  #8240

This PR adds `sign-macos` job to CI workflow, adding a transparent signing step. 

The main reason is to sign repo migrations and avoid issue from https://github.com/ipfs/go-ipfs/issues/8240 globally (outside of Brave). It also provides us with a template if we wish to add signing for other platforms.

### Technical details

This PR splits `make publish` performed in `main.yml` into three stages:
1. `build` executes  `./dockerized make all_dists` and produces any missing binaries in `./releases`
2. `sign-macos` signs and notarizes every new binary from `./releases` and updates `.tar.gz` packages and relevant hashes in-place
3. `publish` 
   * executes `make publish` and adds any new and  signed packages from `./releases` to `DIST_ROOT` 
     (`/ipns/dist.ipfs.io` by default)
   * pins produced CID to ipfs-cluster for persistence

### Flow for adding new dist or version

1. Open a PR against this repo with new entry for one of `./dists/{name}/versions` 
2. Wait for CI to build signed  dist, put it on IPFS and pin updated `DIST_ROOT` to ipfs-cluster
3. Preview link will be published on PR  and printed at the end of `publish` step of CI build
4. If everything looks good, write down the CID  from preview link, and update the DNSlink of dist.ipfs.io

### Content routing optimization

IPFS node running on CI manually pre-connects to ipfs-websites cluster peers, which are known to already have all blocks required during dist.ipfs.io build. This way we skip DHT lookup.

### Website preview on PRs

Every PR now gets a link to website preview at the `dweb.link` IPFS gateway:

> 

## TODO

- [x] sign darwin binaries and put them in the same place as unsigned ones
- [x] ensure all hashes and manifests match signed binaries
- [x] ensure the root CID for updated dist.ipfs.io website is persisted after CI job ends
  - we use `/dnsaddr/ipfs-websites.collab.ipfscluster.io` but [it errors right now](https://github.com/ipfs/distributions/runs/3128450306?check_suite_focus=true#step:14:13)
- [x] gateway link with preview of updated website
- [x] remove fake `DIST_ROOT` and ensure CI is no-op when no new binaries are added
- [x] squash & merge


.github/workflows/main.yml
```diff
@@ -2,40 +2,297 @@ name: CI
...
+# TODO: restore after testing
+#    branches:
+#      - master
+#  pull_request:
+#    branches:
+#      - master
+
+env:
+ # TODO: remove after testing
+  DIST_ROOT: '/ipfs/QmX6J3hVtMF9Y73CKcyLHfgviEfXGFFQABWPXsD17EsBhg' # dist.ipfs.io without ipfs-update@v1.7.1
```
lidel, line 14: This is so we run build against  CID that represents dist.ipfs.io without ipfs-update@v1.7.1 – this demonstrates how additional entry in `dists/ipfs-update/versions`  triggers build of "missing" binary artifacts.

(we will remove this before merging this PR)

.github/workflows/main.yml
```diff
@@ -2,40 +2,287 @@ name: CI
...
+#    branches:
+#      - master
+#  pull_request:
+#    branches:
+#      - master
+
+env:
+ DIST_ROOT: '/ipns/dist.ipfs.io' # content root used for calculating diff to build
+ GO_IPFS_VER: 'v0.9.1'           # go-ipfs daemon used for chunking and applying diff
+ CLUSTER_CTL_VER: 'v0.14.0'      # ipfs-cluster-ctl used for pinning
```
BigLep, line 15: What's the process for updating these?  Do we rely on someone manually updating it?  What's their trigger for doing so?

Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM ubuntu:20.04
 ARG USER_UID
 RUN apt-get update -q && apt-get install -y git curl gnupg jq build-essential gawk zip 
-RUN curl -s https://dist.ipfs.io/go-ipfs/v0.8.0/go-ipfs_v0.8.0_linux-amd64.tar.gz | tar vzx -C /usr/local/bin/ go-ipfs/ipfs --strip-components=1
+RUN curl -s https://dist.ipfs.io/go-ipfs/v0.9.1/go-ipfs_v0.9.1_linux-amd64.tar.gz | tar vzx -C /usr/local/bin/ go-ipfs/ipfs --strip-components=1
```
BigLep, line 4: Similarly, what's the process for updating this?
Can we have a URL like https://dist.ipfs.io/go-ipfs/latest-full-release/go-ipfs_linux-amd64.tar.gz

README.md
```diff
@@ -107,25 +110,27 @@ The optional `sub_package` argument is used to specify a module within a repo.
...
+2. Load the dists website in your browser to make sure everything looks right: `http://localhost:8080/ipfs/<NEW_HASH>`.
+3. Compare `<NEW_HASH>` with the current `dists.ipfs.io` to make sure nothing is amiss: `ipfs object diff /ipns/dist.ipfs.io /ipfs/<NEW_HASH>`
 
 Finally,
 
 1. Commit your changes and make a PR. Specifically, the changes to `dists/<dist>/versions` and `dists/<dist>/current`.
-2. Make a PR with an edit on [protocol/infra](https://github.com/protocol/infra/blob/master/dns/config/dist.ipfs.io.yaml) with the hash you got from `make publish` and a link to the PR above.
+2. Wait for [Github Action](https://github.com/ipfs/distributions/actions/) on your PR to build **signed** binaries. `<NEW_SIGNED_HASH>` will be different than one from local build.
+3. Make a PR with an edit on [protocol/infra](https://github.com/protocol/infra/blob/master/dns/config/dist.ipfs.io.yaml) with `<NEW_SIGNED_HASH>` you got from the Github Action output and a link to the PR above.
+   - TODO: this step will be automated in the future.
```
BigLep, line 133: Can we link to the issue that tracks this?

build-go.sh
```diff
@@ -9,8 +9,9 @@ export GOPATH
 # Always use go modules
 export GO111MODULE=on
 
-DIST_PATH=${DIST_PATH:-/ipns/dist.ipfs.io}
-DIST_PATH=$(ipfs resolve "$DIST_PATH")
+# Content path to use when looking for pre-existing release data
+DIST_ROOT=${DIST_ROOT:-/ipns/dist.ipfs.io}
+DIST_ROOT=$(ipfs resolve "$DIST_ROOT")
```
BigLep, line 14: nit: Seems odd in one script to overwrite a variable like this.  Maybe do as one line?

.github/workflows/main.yml
```diff
@@ -2,40 +2,287 @@ name: CI
...
+          name: tmp
+          path: ./tmp/
+          retention-days: 1
+      - name: Update changed binaries in ./releases
+        run: |
+          for NEW_DIR in ./releases/*/v*; do
+            (! test -d "$NEW_DIR") && continue
+            DIST_VERSION=$(basename "$NEW_DIR")
+            DIST_NAME=$(basename $(dirname "$NEW_DIR"))
+            for arch in "amd64" "arm64"; do
```
BigLep, line 155: Is it normal to have big shell scripts inlined like this in GitHub workflow?  Or would it be better to factor it out?

(I don't know  conventions here).

Also, do we have a "3 finger rule" around bash (i.e., once bash code is more than 3 fingers of lines on the terminal, we use a different language that is more developer friendly)? (We don't need to introduce that here - just curious - feel free to disregard()

BigLep (commented):
Thanks for doing this!  I'll leave it for others to approve.

.github/workflows/main.yml
```diff
@@ -2,40 +2,287 @@ name: CI
...
+          API_PARAMS=$(jq --monochrome-output --null-input \
+            --arg state "success" \
+            --arg target_url "$PREVIEW_URL" \
+            --arg description "Preview on IPFS" \
+            --arg context "IPFS" \
+            '{ state: $state, target_url: $target_url, description: $description, context: $context }' )
+          curl --output /dev/null --silent --show-error \
+            -X POST -H 'Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' -H 'Content-Type: application/json' \
+            --data "$API_PARAMS" 'https://api.github.com/repos/ipfs/distributions/statuses/${{ github.sha }}'
+          echo "Pinned to IPFS - $PREVIEW_URL"
```
olizilla, line 288: This mechanism for adding a status result to a PRs checks list stopped working for me in ipfs-dns-deploy. I tried to fix it, but could never get the status to appear again on the PR... what did you change to make it work here!?

.github/workflows/main.yml
olizilla: I would support such a thing, but i don't believe there are any existing conventions for it.

.github/workflows/main.yml
lidel: I think those are low importance: could be kept as-is until someone cares to update or there is a breaking change that forces version bump.

Having static versions of build tools makes the builds more robust and reproducible, but we could also switch to always run on the latest version for dogfooding – don't feel strongly either way, but leaning towards more reproducibility.

.github/workflows/main.yml
lidel: @olizilla Hm.. not sure, could be one of below:
* copied it from https://github.com/ipfs-shipyard/ipfs-github-action/blob/182f36dcd28a0df011ce1ee9b004a1ca7fa7c720/entrypoint.sh#L11-L32 and tweaked quotes and renamed some fields.
* stopped installing `jq`  via snap and use the version shipped with `ubuntu-latest`. 
* use `GITHUB_TOKEN` secret directly, and not via env
  * I would guess this is most likely the cause. Github hardened some things around their tokens this year, [enabling orgs to tweak defaults or limit the scope](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/) (maybe PL changed something?). I've read that some github actions stopped working with regular `GITHUB_TOKEN` and require creation of a [a PAT (Personal Access Token)](https://docs.github.com/en/enterprise-server@3.1/github/authenticating-to-github/keeping-your-account-and-data-secure/creating-a-personal-access-token) (I think only official one under `actions/*`  continue to work "old way", but also did not dig into this dep enough to understand what exactly changed and why)

README.md
lidel: Added https://github.com/ipfs/distributions/issues/372 in f115f93

Dockerfile
lidel: Good catch. Simplified in 7d222b5 which makes sure HTTP client here uses same version as `GO_IPFS_VER` – just in case we have breaking change  in any of used APIs.

.github/workflows/main.yml
lidel: I feel shell scripts are the right tool for the job in CI workflows, but this could be my personal bias :)
Considered extracting those longer blocks to `.sh` files in  `./scripts/ci/` but it felt harder to read and reason about – left them inlined.

We can extract them after this + nightlies lands, but suggest doing that in a separate PR, to make reviews at hand easier by having everything in one place, removing need for jumps between files.

build-go.sh
lidel: Simplified in 9acff7e :+1:

.github/workflows/main.yml
lidel: Update: after reviewing #369 I believe I need to extract those bash blocks into reusable scripts to land everything anyway :) (want to run them in both `main.yml` for release / PR builds and `nightly.yml` for daily matrix build).
Stay tunned.

lidel: Done. Extracted scripts are in `./scripts/ci` and will be shared with nightly  setup.

scripts/ci/github-preview-link.sh
```diff
@@ -0,0 +1,14 @@
...
+PREVIEW_URL="https://dweb.link$CONTENT_PATH"
+API_PARAMS=$(jq --monochrome-output --null-input \
+    --arg state "success" \
+    --arg target_url "$PREVIEW_URL" \
+    --arg description "Preview updated website on IPFS" \
+    --arg context "Preview is ready" \
+    '{ state: $state, target_url: $target_url, description: $description, context: $context }' )
+curl --output /dev/null --silent --show-error \
+    -X POST -H "Authorization: Bearer $GITHUB_TOKEN" -H 'Content-Type: application/json' \
+    --data "$API_PARAMS" 'https://api.github.com/repos/ipfs/distributions/statuses/${GIT_REVISION}'
```
lidel, line 13: :information_source: we could replace this script with JS call and  [`actions/github-script`](https://github.com/actions/github-script) but I don't want to sink any more time into refactoring this PR ;)

dockerized
```diff
@@ -5,8 +5,11 @@ set -euxo pipefail
 docker pull ubuntu:20.04
 
 # CACHEBUST means this will apply the updates once a day
-docker build . -t distributions --build-arg CACHEBUST=`date --iso-8601=date` --build-arg USER_UID=$(id -u "$USER")
+docker build . -t distributions \
+    --build-arg CACHEBUST=`date --iso-8601=date` \
+    --build-arg USER_UID=$(id -u "$USER") \
+    --build-arg GO_IPFS_VER=${GO_IPFS_VER:-$(curl -s https://dist.ipfs.io/go-ipfs/versions | tail -n 1)} # match http api client version on CI
```
lidel, line 11: :information_source: this controls the version of the CLI client _inside_ `./dockerized` build, which talks to go-ipfs running _outside_ on the host. Here, we ensure it uses the same version (with fallback to the latest one if `GO_IPFS_VER` is undefined).

scripts/ci/pin-to-cluster.sh
```diff
@@ -0,0 +1,28 @@
...
+echo "::endgroup::"
+
+echo "::group::waiting until pinned"
+    while true; do
+    ipfs-cluster-ctl --enc=json \
+        --host "/dnsaddr/ipfs-websites.collab.ipfscluster.io" \
+        --basic-auth "${CLUSTER_USER}:${CLUSTER_PASSWORD}" \
+        status "$PIN_CID" | tee cluster-pin-status
+    if [[ $(jq '.peer_map[].status' cluster-pin-status | grep '"pinned"' | wc -l) -ge 2 ]]; then
+        echo "Got 2 pin confirmations, finishing the workflow"
```
lidel, line 21: :information_source: to make CI faster and avoid hanging forever if parts of cluster are out of sync, cluster pinning is considered "done" after we have 2 copies. 

Here, I do it manually in userland, but filled https://github.com/ipfs/ipfs-cluster/issues/1427 to add this sort of thing to ipfs-cluster-ctl CLI tool.

scripts/ci/setup-ipfs.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+set -e
+
+echo "::group::Install go-ipfs and ipfs-cluster-ctl"
+    curl -s https://dist.ipfs.io/go-ipfs/${GO_IPFS_VER}/go-ipfs_${GO_IPFS_VER}_linux-amd64.tar.gz | sudo tar vzx -C /usr/local/bin/ go-ipfs/ipfs --strip-components=1
+    curl -s https://dist.ipfs.io/ipfs-cluster-ctl/${CLUSTER_CTL_VER}/ipfs-cluster-ctl_${CLUSTER_CTL_VER}_linux-amd64.tar.gz | sudo tar vzx -C /usr/local/bin/ ipfs-cluster-ctl/ipfs-cluster-ctl --strip-components=1
+echo "::endgroup::"
+
+# fix resolv - DNS provided by Github is unreliable for DNSLik/dnsaddr
+sudo sed -i -e 's/nameserver 127.0.0.*/nameserver 1.1.1.1/g' /etc/resolv.conf
```
lidel, line 10: :information_source: This was pretty annoying. Entire build failed because Githubs DNS proxying cache at `127.0.0.*`  randomly timeouted while resolving `_dnslink` or `_dnsaddr`. Switching to Cloudflare here make the problem go away.

scripts/ci/sign-new-macos-releases.sh
```diff
@@ -0,0 +1,92 @@
...
+    # ./tmp/${DIST_NAME}_${DIST_VERSION}_${arch}-unsigned/
+    for NEW_DIR in ./releases/*/*; do
+        (! test -d "$NEW_DIR") && continue
+        DIST_VERSION=$(basename "$NEW_DIR")
+        DIST_NAME=$(basename $(dirname "$NEW_DIR"))
+        for arch in "amd64" "arm64"; do
+            EXECUTABLES=$(jq -nc '$ARGS.positional' --args $(find "./tmp/${DIST_NAME}_${DIST_VERSION}_${arch}-unsigned/" -perm +111 -type f -print))
+            echo "{
+                \"source\" : $EXECUTABLES,
+                \"bundle_id\" : \"io.ipfs.dist.${DIST_NAME}\",
```
lidel, line 30: :information_source: I registered the wildcard `bundle_id=io.ipfs.dist.*`  at Apple, which enables us to have each package "isolated" from each other. This is important in case Apple decides to block something from running on macOS – we don't have all egs in one basket.

.github/workflows/main.yml
```diff
@@ -2,40 +2,131 @@ name: CI
...
+        continue-on-error: true # skip if no releases
+      - name: List ./releases before
+        run: ls -Rhl ./releases || echo "No ./releases"
+      - name: Install gon via HomeBrew for code signing and app notarization
+        run: |
+          brew tap mitchellh/gon
+          brew install ipfs coreutils gawk gnu-sed jq mitchellh/gon/gon
+          ipfs init --profile test # needed for calculating NEW_CID later
+      - name: Import Keychain Certs
+        uses: apple-actions/import-codesign-certs@253ddeeac23f2bdad1646faac5c8c2832e800071 # v1@2020-02-03
```
lidel, line 72: :information_source: :policeman:  I hardcoded specific revision because this is a third-party action and in theory someone could swap-out git tag and steal our signing keys.

.github/workflows/main.yml
```diff
@@ -2,40 +2,131 @@ name: CI
...
+        with:
+          name: releases-unsigned-diff
+          path: releases
+        continue-on-error: true # skip if no releases
+      - name: List ./releases before
+        run: ls -Rhl ./releases || echo "No ./releases"
+      - name: Install gon via HomeBrew for code signing and app notarization
+        run: |
+          brew tap mitchellh/gon
+          brew install ipfs coreutils gawk gnu-sed jq mitchellh/gon/gon
```
lidel, line 69: :information_source: [gon](https://github.com/mitchellh/gon) package is used by Terraform project, so multiple eyes are looking at this

lidel (commented):
Ready for review, aiming for merging in the beginning of the next week.

README.md
```diff
@@ -83,13 +87,20 @@ Run:
...
 ```sh
 > ./dist.sh add-version fs-repo-99-to-100 v1.0.1
 ```
 
+To produce a signed, **official build** for use in DNSLink at `dist.ipfs.io`:
+
+1. Run `./dist.sh add-version` locally.
+2. Commit created changes to `dists/<dist>` and open a PR against `ipfs/distributions`.
+3. Wait for Github Action to finish PR build. It runs `./dockerized` build, then signs macOS binaries and spits out updated root CID at the end.
+4. If everything looks good, write down the CID from the preview link on the PR, and update the DNSlink at `dist.ipfs.io`.
```
lidel, line 102: :information_source:   I believe this is the TLDR of this entire PR :)

README.md
```diff
@@ -194,6 +207,16 @@ So for example, if we had `<dist>` `go-ipfs` and `fs-repo-migrations`, we might
...
 
+### Notes on reproducible builds
+
+Running `./dockerized make publish` will produce binaries using the same
+runtime as CI. The main difference between local build and official CI one is
+signing step on platforms such as `darwin` (macOS).
+
+Signatures are attached at the end of macOS binaries, which means
+`*_darwin-*.tar.gz` produced by CI will have additional bytes when compared
+with local build.
```
aschmahmann, line 218: TODO (post PR): Let's add some references to tools for how to do the binary stripping so people don't have to search for it.

aschmahmann (approved):
LGTM, although I mostly reviewed the README. We may learn more as we start using it for nightlies though ❤️

dockerized
BigLep: Should this comment live in the source file?

scripts/ci/pin-to-cluster.sh
BigLep: Similar to above: when do we decide to put this in the PR vs. the code itself?  Do folks tend to do a git blame and look for comments to find more context?

BigLep (approved):
Great work!
Only comments were around inlining comments PR into the code for discoverability.  Your call.

dockerized
lidel: I believe it is good to have it here in case someone else tried to modify those scripts in the future   – it explains why we install go-ipfs twice (daemon + client in docker)

scripts/ci/pin-to-cluster.sh
lidel: I often use git blame,  but in this specific case those comments are provided mostly to make PR review easier.

README.md
lidel: Filled https://github.com/ipfs/distributions/issues/377 so we dont forget ;)

scripts/ci/pin-to-cluster.sh
BigLep: Ok.  I'll defer to the team for the standard.  I was bringing this up because it was hard for me to imagine a case where a comment is useful during PR where it isn't also useful 6 months in the future when looking at the code with fresh eyes.  Anyways, sounds good.

lidel:
As agreed during stewards call, I'm starting the merge dance starting with this PR

PR #369, opened 2021-07-22
feat: create nightly builds

### LINKED ISSUE #363: Create nightly builds
> cc / supersede https://github.com/ipfs/distributions/issues/77 & https://github.com/ipfs/distributions/issues/78

## Motivation

We want to have a setup that constantly proves our release pipeline works as expected. Producing a nightly builds for all dist packages is the best way for identifying issues early and minimizing risk of  delays of Stable and RC releases.

This is the main reason for doing this. 
Providing nightly builds to the community is a nice side effect, but not something we will explicitly support or advertise anywhere.

## Architecture

- We want a daily confirmation that everything can be built, so we want to cut a new release without worrying about CI.
- Nightly builds will be automated via Github Actions, and will reuse as much of the regular release pipeline, as possible.
-  Would be nice to have a separate website at `nightly.dist.ipfs.io` will provide N days of nightly builds, but keeping it limited to  github actions is enough for MVP
- Pin artifacts to website cluster for persistence 

### Rough plan 
- [x] extend `./dist.sh add-version <dist> <version>` with suport for a special " `nightly`" version
  - [x] `nightly` should build the latest revision from the _default_ branch  (don't  assume it is `master`)
  - [x] artifact names should have `nightly-2021-07-16-9599ad5` instead of regular versions like  `v0.9.0`  (build date and first 7 characters of  git revision used as the source)
  - [x] keep only N nightlies around (eg. 7)
  - [x] <del>use `/ipns/nightly.dist.ipfs.io` or </del> `/ipns/dist.ipfs.io` as the base root (update: we want nightly test that adds patch on top of existing website, so original  `/ipns/dist.ipfs.io`  has to be used)
  - [ ] idempotent within each day: skip build if the revision was already built for today
- [ ] <del>set up DNS infra for  `nightly.dist.ipfs.io`, add `DNSIMPLE_TOKEN` to `ipfs/distributions` repo  to enable update via CI ([prior art](https://github.com/ipfs/ipfs-webui/blob/e546203b5276698eedcdb07d0c90fec9cbb607cd/.github/workflows/ci.yml#L134))</del> 
- [x]  add a new cron workflow to Actions that builds a `nightly` every day. 
  - [x] Create `.github/workflows/nightly.yml` 
    - [x] Run `make publish` pipeline  just like `main.yml`, but find a way to build nightly for every package defined in `./dists` directory
      - [x]  `echo "nightly_2021-07-16" >> dists/go-ipfs/versions` and then make publish producing nightly binaries in `./releases`
           - this has cool property of  enabling us to `make all_dists` (build all binaries) for each nightly in separate, parallel  job, and then merge  `./dists` and` ./releases`  before running make publish in final step (generating website and putting everything on IPFS as a patch on top of `DIST_ROOT`)
      - [x] if any package failed in the middle of the pipeline, the whole nightly job should fail (but we should set the "failed" flag at the very end, so we see all packages that are broken, not only the first one)
      - [x]  leverage `DIST_ROOT=/ipfs/Qm..` for starting with an empty or customised root for nightlies without waiting for dnslink at `nightly.dist.ipfs.io`

iand:
Closes #363

### High level summary

- This PR  adds a cron job that runs every day and builds nightly releases for each package name in `./dists`
- `nightly.yml` workflow is built on top of modernized `main.yml` (#367)  and runs the same builds scripts
  - building 23+ packages takes time, so we build them separately, in parallel – each one having own isolated output proving it can be built end-to-end
  - global status can be quickly inspected on https://github.com/ipfs/distributions/actions/workflows/nightly.yml and clicking on the latest run allows further inspection https://github.com/ipfs/distributions/actions/runs/1077161039
  - :point_right:  If one of the packages fails to build, workflow will create a new issue (or comment on existing open one) – examples: https://github.com/ipfs/distributions/issues/374 & https://github.com/ipfs/distributions/issues/375

### Build scripts changes 

 - `./dist.sh add-version <dist> nightly` will add a nightly version to the versions file using the current date (@lidel is this still useful?)
 - `make nightly` add a nightly version to every dist versions file using the current date
    - one can also `cd dists/<package>  && make nightly` to add it only to specific package - we use it on CI to run builds in paralllel (@lidel)
 - `make dist` will build nightly versions by checking out the latest version of the default branch and producing a build artifact named using the pattern `nightly-yyyy-mm-dd-hash` where hash is the starting 7 (at least) characters of the commit hash. It may be more than 7 if a shorter length would be ambiguous.
 - `make all_dists` is unchanged but will build a nightly version in the course of running `make dist` above
 - It's not expected that the `nightly-yyyy-mm-dd` version will be committed to the repo. It's an ephemeral version that is generated to produce a nightly build.
 - The final `versions` file is normally added to ipfs my `make publish`. This will now include the most recent nightly version. This allows the build system to ignore it on subsequent runs until a new day is reached.

### TODO
- [x] wire up ability to build `nightly-*` versions using existing `make publish` pipeline (@iand)
- [x] add publish step to nightly.yml (@iand)
- [x]  do not pollute `./nightly` (@lidel)
- [x] ensure build does not use commits older than the nightly threshold (@lidel)
   TODO: explore checking out the last commit in the default branch before arbitrary  date:  `git rev-list -1 --first-parent --before="2021-07-27" "$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')"` – makes nightlies more deterministic (at the revision level)
- [x] fix vtag releases (like `fs-repo-migrations/fs-repo-0-to-1`) (@lidel)
- [x] share build scripts with `main.yml` (@lidel)
- [ ] refactor to include signing (or make part of `main.yml`
  - extracted to https://github.com/ipfs/distributions/issues/378


common.mk
```diff
@@ -7,15 +7,23 @@ relpath = $(dir $(lastword $(MAKEFILE_LIST)))
 # Default values
 distname ?= $(notdir ${CURDIR})
 releases ?= $(relpath)releases/${distname}
+versions ?= versions
+
+nightlyVer = nightly-$(shell date '+%Y-%m-%d')
```
lidel, line 12: In case Actions use worker with different timezone, lets ensure UTC
```suggestion
nightlyVer = nightly-$(shell date -u '+%Y-%m-%d')
```

dist.sh
```diff
@@ -74,6 +74,7 @@ case $1 in
 
 		case "$nvers" in
 			*-*) echo "WARNING: not marking pre-release $dist $nvers as the current version." ;;
+			nightly) nvers=$nvers-$(date '+%Y-%m-%d') ;;
```
lidel, line 77:
```suggestion
			nightly) nvers=$nvers-$(date -u '+%Y-%m-%d') ;;
```

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,26 @@
+name: Nightly
+
+on:
+  schedule:
+    - cron: '30/15 12 * * *' # UTC
```
iand, line 5: The minute directive here is temporary for testing

iand:
This doesn't handle any specific publishing logic for nightlies. I haven't tested but I expect that one nightly build for each day will be published. 

How do we want to handle clean up of old nightlies? What is the best mechanism for keeping track of how many nightlies have been built so we can restrict to a given number, e.g. 7.

lidel:
Number of nightlies per package can be calculated by  `grep nightly ./dists/{name}/versions | wc -l` 

I imagine dropping of the oldest nightlies over the limit could happen in `scripts/patch.js` (executed during `make publish`) after everything is patched, but  right before returning the final CID.

iand:
> Number of nightlies per package can be calculated by `grep nightly ./dists/{name}/versions | wc -l`

No, since `versions` will contain only the most recent nightly since the nightly CI does not commit updates to the versions file back to the repo. See 4th bullet in the pr description. 

I think the simplest place to do it is in `build-go.sh` as an additional step (5) in the existing sequence:

 1. check out repo ,`dists/{name}/versions` just contains tagged versions
 2. append nightly version to `dists/{name}/versions`
 3. compare with ipfs `dists/{name}/versions` and build anything new 
 4. copy `dists/{name}/versions` to `releases/{name}/versions`
 5. **append `nightly` builds from ipfs `dists/{name}/versions` to `releases/{name}/versions` (keeping 7)** 
 6. publish `releases/{name}/versions` as ipfs `dists/{name}/versions`  which will now contain tagged versions and 7 nightly


> 
> I imagine dropping of the oldest nightlies over the limit could happen in `scripts/patch.js` (executed during `make publish`) after everything is patched, but right before returning the final CID.

lidel (commented):
Works as expected, thank you, @iand! :raised_hands: 

---- 

I will take this over and tweak it a bit to be compatible with final approach from #367
(will comment when ready)

lidel:
- This is ready for review, but should be merged after #367
- `nightly.yml` workflow is built on top of the modernized `main.yml` (#367) and runs the same builds scripts
  - it does not do signing of produced nighties yet because of [macOS worker limits](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits), but we can add it soon
  - building 23+ packages takes time, so we build them separately, in parallel – each one having own isolated output prooving it can be built end-to-end
  - global status can be quickly inspected on https://github.com/ipfs/distributions/actions/workflows/nightly.yml and clicking on the latest run allows further inspection https://github.com/ipfs/distributions/actions/runs/1077161039
  - :point_right: If one of the packages fails to build, workflow will create a new issue (or comment on existing open one) – examples: #374, #375

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,104 @@
+name: Nightly
+
+on:
+  push:
+  ## Allow manual invocation
+  #workflow_dispatch:
+  #schedule:
+  #  - cron: '0 5 * * *' # UTC
```
lidel, line 8: :information_source:  we will remove `push` and restore `schedule` for cron trigger + `workflow_dispatch` for running nightly builds on-demand

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,104 @@
...
+      - name: Setup IPFS
+        run: ./scripts/ci/setup-ipfs.sh
+        env:
+          CLUSTER_USER: ${{ secrets.CLUSTER_USER }}
+          CLUSTER_PASSWORD: ${{ secrets.CLUSTER_PASSWORD }}
+        timeout-minutes: 5
+      - run: cd ./dists/${{ matrix.dist_name }} && make nightly
+      - run: ./dockerized make publish
+      - name: Create issue if build failed
+        uses: actions/github-script@v4
```
lidel, line 48: :information_source: Third-party turn-key solutions exists, but I decided to give `actions/github-script` a try.  It is an official action provided by GitHub, provides JS porcelain for running arbitrary API calls. 

Debugging is a bit painful, but being written in JS could make it more dev friendly than bash+curl, allowing more people to maintain/extend.  `actions/github-script` is extremely powerful, only caveat is that it needs a [personal access token](https://docs.github.com/en/github/authenticating-to-github/keeping-your-account-and-data-secure/creating-a-personal-access-token) to do anything outside the current repo.

We may leverage it even more for developer flow automation tailored to our needs. For example, it could automatically create/label issues  and add them to specific column in specific project.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,104 @@
...
+        run: git status && ls -Rhl ./releases
+      - name: Read CID of updated DAG
+        id: cid-reader
+        run: echo "::set-output name=CID::$(tail -1 ./versions)"
+      - name: Pin new website to ipfs-websites.collab.ipfscluster.io
+        run: ./scripts/ci/pin-to-cluster.sh
+        env:
+          PIN_CID: ${{ steps.cid-reader.outputs.CID }}
+          PIN_NAME: "ipfs/distributions/nightly/${{ matrix.dist_name }}"
+          PIN_ADD_EXTRA_ARGS: "--expire-in 168h"
```
lidel, line 94: :information_source: this pins nightly artifacts only for a week

lidel (approved):
Ready for review.

BigLep (commented):
Great work!
One small idea is to consider the comments you  left in the PR to inline them into the code for discoverability.

PR #371, opened 2021-07-22
Don't silence Hugo

thattommyhall:
This was masking errors when running from Make

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "webpack:watch": "webpack --watch-stdin",
     "less:watch": "onchange 'site/assets/less/**/*.less' -- npm run less:build",
     "less:build": "lessc site/assets/less/main.less --clean-css site/static/site.css",
-    "build:hugo": "hugo --quiet -s site -d ../releases",
+    "build:hugo": "hugo -s site -d ../releases",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -64,4 +64,4 @@
     "webpack": "^4.41.0",
     "webpack-cli": "^3.3.9"
   }
-}
+}
```
aschmahmann, line 67: I think these files seem to like newlines at the end for some reason

package.json
```diff
@@ -12,7 +12,7 @@
     "webpack:watch": "webpack --watch-stdin",
     "less:watch": "onchange 'site/assets/less/**/*.less' -- npm run less:build",
     "less:build": "lessc site/assets/less/main.less --clean-css site/static/site.css",
-    "build:hugo": "hugo --quiet -s site -d ../releases",
```
aschmahmann, line 15: Seems reasonable to me, although I know almost nothing about Hugo so I'll defer to @lidel to double check me.

hugomrdias:
Thank you 😆

olizilla:
I silenced hugo cos he talks to much. but masking errors is unhelpful, so go for it.

PR #376, opened 2021-07-30
fix(release): assure new release version is greater than the current one

### LINKED ISSUE #319: dist.sh should check that the new version is newer than the current version
A note-to-self to make dist.sh verify that the new-version is in fact newer in semver terms than the current version... we avoid setting release candidates as current, but we dont guard against setting older versions as current. Minor but feels like a foot gun.

https://github.com/ipfs/distributions/blob/211588d697902572642a53d20b2ff5d121ecd1b9/dist.sh#L58-L63

AlexxNica:
Fixes #319

### DIFF
--- a/dist.sh
+++ b/dist.sh
@@ -15,6 +15,29 @@ sedEscapeArg() {
 	echo "$@" | sed 's/\//\\\//g'
 }
 
+# Adapted from https://stackoverflow.com/a/44660519/702738
+compare_version() {
+	if [[ $1 == $2 ]]; then
+		return 0
+	fi
+	local IFS=.
+	local i a=(${1%%[^0-9.]*}) b=(${2%%[^0-9.]*})
+	local arem=${1#${1%%[^0-9.]*}} brem=${2#${2%%[^0-9.]*}}
+	for ((i=0; i<${#a[@]} || i<${#b[@]}; i++)); do
+		if ((10#${a[i]:-0} < 10#${b[i]:-0})); then
+				return 1
+		elif ((10#${a[i]:-0} > 10#${b[i]:-0})); then
+				return 0
+		fi
+	done
+	if [ "$arem" '<' "$brem" ]; then
+		return 1
+	elif [ "$arem" '>' "$brem" ]; then
+		return 0
+	fi
+	return 1
+}
+
 case $1 in
 	new-go-dist)
 		name="$2"
@@ -58,7 +81,7 @@ case $1 in
 
 		# Create vtag file that contains version tag prefix
 		if [ "$tag_prefix" != "." ]; then
-			echo "$tag_prefix" > "dists/${name}/vtag"
+			echo "$tag_prefix" > "dists/$name/vtag"
 		fi
 
 		echo "distribution $name created successfully! To start build: make $name"
@@ -72,9 +95,20 @@ case $1 in
 			exit 1
 		fi
 
+		# Test if the new version is greater than the current version.
+		test_version() {
+			current_version=$([ -f "dists/$dist/current" ] && cat "dists/$dist/current" | sed "s/v//")
+			new_version=$(echo $nvers | sed "s/v//")
+
+			if compare_version "$current_version" "$new_version"; then
+				echo "ERROR: The version provided ($new_version) has to be greater than the current release ($current_version)."
+				exit 1
+			fi
+		}
+
 		case "$nvers" in
-			*-*) echo "WARNING: not marking pre-release $dist $nvers as the current version." ;;
-			*) echo "$nvers" > "dists/$dist/current" ;;
+			*-*) test_version && echo "WARNING: not marking pre-release $dist $nvers as the current version." ;;
+			*) test_version && echo "$nvers" > "dists/$dist/current" ;;
 		esac
 
 		echo "$nvers" >> "dists/$dist/versions"


dist.sh
```diff
@@ -72,9 +95,20 @@ case $1 in
...
 
+		# Test if the new version is greater than the current version.
+		test_version() {
+			current_version=$([ -f "dists/$dist/current" ] && cat "dists/$dist/current" | sed "s/v//")
+			new_version=$(echo $nvers | sed "s/v//")
+
+			if compare_version "$current_version" "$new_version"; then
+				echo "ERROR: The version provided ($new_version) has to be greater than the current release ($current_version)."
+				exit 1
+			fi
```
gammazero, line 106: This does not appear to catch the problem when:
```sh
current_version="2.3-rc1"                                                                                      
new_version="2.3"                                                                                              
```
I get:
```sh
ERROR: The version provided (2.3) has to be greater than the current release (2.3-rc1).
```
The version "2.3-rc1" should, semver-wise, be less than "2.3".

Running a tool like [semver-cli](https://github.com/davidrjonas/semver-cli) and does catch this:
```sh
> semver-cli greater -v "2.3-rc1" "2.3"; echo "returned: $?"
2.3
returned: 1
> semver-cli greater -v "2.3" "2.3-rc1"; echo "returned: $?"
2.3
returned: 0
```

Consider using [semver-cli](https://github.com/davidrjonas/semver-cli) and changing the logic above to:
```sh
if ! semver-cli greater "$new_version" "$current_version"; then
	echo "ERROR: The version provided ($new_version) has to be greater than the current release ($current_vers\
ion)."
	exit 1                                                                      
fi
```

gammazero (changes requested):
There appears to be a problem with the logic when comparing rc versions.

Please be sure to test with all combinations of semver values.

dist.sh
AlexxNica: @gammazero, that's because `2.3-rc1` is not compliant with the SemVer specification, as noted on [Item 2 of the specification](https://semver.org/#spec-item-2):

> A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and MUST NOT contain leading zeroes. X is the major version, Y is the minor version, and Z is the patch version. Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.

Thus, to be correctly formatted, the version must be separated by three dots.

There are some [useful tips on the SemVer's FAQ](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string) that I use myself:

> Q: Is there a suggested regular expression (RegEx) to check a SemVer string?
>
> A: There are two. One with named groups for those systems that support them (PCRE [Perl Compatible Regular Expressions, i.e. Perl, PHP and R], Python and Go).
> See: https://regex101.com/r/Ly7O1x/3/
>
> `^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`
>
> And one with numbered capture groups instead (so cg1 = major, cg2 = minor, cg3 = patch, cg4 = prerelease and cg5 = buildmetadata) that is compatible with ECMA Script (JavaScript), PCRE (Perl Compatible Regular Expressions, i.e. Perl, PHP and R), Python and Go.
> See: https://regex101.com/r/vkijKf/1/
>
> `^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`

gammazero: Thank you for the detailed explanation.  I took an example directly from the specification (item 11-3)
> When major, minor, and patch are equal, a pre-release version has lower precedence than a normal version:
Example: 1.0.0-alpha < 1.0.0.

The code provided does not catch that "1.0.0-alpha" is less than "1.0.0"

BigLep:
@AlexxNica: are you able to incorporate comments so we an get this merged?

AlexxNica:
> @AlexxNica: are you able to incorporate comments so we an get this merged?

@BigLep absolutely! Will get that done ASAP.

BigLep:
Awesome.  Please set the status to "In Review" when it's ready to be looked at again.  Thanks!

BigLep:
@AlexxNica : just checking in.  Are you able to incorporate comments?

AlexxNica:
@BigLep 

Sorry for the delay!

I'd suggest we integrate any reliable package from npm that already handles everything SemVer-related and call it from the bash script to validate versions.

I couldn't easily find a ready-to-use and reliable way of checking and comparing SemVer versions directly within bash, so migrating to a node script of some sort would be the best option unless we have the need to keep this logic within a bash script.

dist.sh
AlexxNica: @gammazero, thank you for your review! You're right! And thanks for commenting on the StackOverflow answer related to this!

BigLep:
@AlexxNica : I agree that it doesn't seem like a good idea to implementing this kind of thing in Bash - ugh.  I dont' have experience with this repo, but if there's an npm package you have in mind for pulling in, feel free to do so.  Thanks.

AlexxNica:
@BigLep, Yea, implementing that in Bash is not a pretty thing.. 😰

As for the npm package, I'd use the battle-tested package used by node itself: https://github.com/npm/node-semver.

I'll put this in my backlog and try getting it done within the next two weeks.

AlexxNica:
Update: having to push this to probably the end of the month due to a project launch we'll have between this and next week.

In the meantime, I've updated the PR description with a new checklist of to-dos.

cc @BigLep

PR #381, opened 2021-08-04
feat: fs-repo-migrations v2.0.2 (macos signed)

lidel:
This PR adds no-op version bump (patch release) for every `fs-repo-*` packaged  to build binaries using the new CI
pipeline that provides macOS notarization and enables us to  fix https://github.com/ipfs/go-ipfs/issues/8240 in go-ipfs 0.10.

It respects darwin archs defined in `dists/{dist_name}/build_matrix` and only signs existing 12 packages.

Most likely will produce a humongous diff, so below is a shortened list of added releases:

```
./releases
├── fs-repo-0-to-1
│  ├── v1.0.1
│  └── versions
├── fs-repo-1-to-2
│  ├── v1.0.1
│  └── versions
├── fs-repo-2-to-3
│  ├── v1.0.1
│  └── versions
├── fs-repo-3-to-4
│  ├── v1.0.1
│  └── versions
├── fs-repo-4-to-5
│  ├── v1.0.1
│  └── versions
├── fs-repo-5-to-6
│  ├── v1.0.1
│  └── versions
├── fs-repo-6-to-7
│  ├── v1.0.1
│  └── versions
├── fs-repo-7-to-8
│  ├── v1.0.1
│  └── versions
├── fs-repo-8-to-9
│  ├── v1.0.1
│  └── versions
├── fs-repo-9-to-10
│  ├── v1.0.1
│  └── versions
├── fs-repo-10-to-11
│  ├── v1.0.1
│  └── versions
└── fs-repo-migrations
   ├── v2.0.2
   └── versions
```


gammazero (approved):
From what I can tell this looks good.  There is one thing I noticed...

This is building for `darwin-arm64` for only `fs-repo-7-to-8` and later.  I think that was previously a conscious decision because did not think anyone on a new Mac would ever need that old of migration.  However, older migrations are now getting a version bump for macos signing, but are still not being built for `darwin-arm64`.   I just wanted to point out that slight inconsistency in case you think it matters.

lidel:
Ack. My assumption is that we want to keep build targets intact – there is no point in building older migration for architecture such as arm64  if we did not provide go-ipfs for it.

gammazero:
@lidel ... unless I move my drive, that I keep all IPFS data on, to new computer because old computer stopped working

lidel:
Unlikely (given how old the olderst migrations are), but it also does not cost us anything to add arm64 for them, so added in 4ccc3a1 :)

gammazero (approved):

lidel:
Fixed a bug I introduced during one of rebases. Fix is in https://github.com/ipfs/distributions/pull/381/commits/a8f62b66521e439625acdd1bbda7623d015aac66

TLDR: we were not appending the new versions to `./releases/<dist>/versions` during the build (was happening only for nightlies), so go-ipfs was not picking up new versions of migraitons correctly. Rebuilt version below will have correct `versions`

lidel:
`IPFS_DIST_PATH=/ipfs/QmUyaK98GunkSCUuCp3dQiakbVT4CNF2NCUMioPDgkLrAZ` did not pass the smoke test:

`fs-repo-10-to-11` binary is signed and notarized ([we got the confirmation from apple's notary service](https://github.com/ipfs/distributions/runs/3264808802?check_suite_focus=true#step:8:555)), however when migration is executed within go-ipfs spawned by Brave, it still shows the error:

> vtQC6sJ
> wfCj33T

Will look into this, need to understand if the signed and notarized binary is present in the produced tree.

lidel:
Found and fixed a bug related to in https://github.com/ipfs/distributions/pull/381/commits/5e069a47e68c767639d44e5c6c50f1d2833dd648:

> This fixes a bug where only the very first dist got signed due to relative pathing inside of .sh script that applied signed updates on top of existing packages.

Will re-test when new CID is generated and posted below.

lidel:
:checkered_flag:  Confirmed  `IPFS_DIST_PATH=/ipfs/QmP7tLxzhLU1KauTRX3jkVkF93pCv4skcceyUYMhf4AKJR` works fine:
XS2BOv6

Will open PR against go-ipfs shortly.

lidel:
PR that switches go-ipfs to signed migrations: https://github.com/ipfs/go-ipfs/pull/8333
@aschmahmann ok to merge this one?

lidel:
There were no concerns raised during sync today – merging this to unblock ipfs/go-ipfs#8333

Issue #372, opened 2021-07-27
Should we update DNSLink on merge to master?


lidel:
Current workflow assumes DNSLink will be updated manually by filling a PR against [this `yml` in protocol/infra](https://github.com/protocol/infra/blob/master/dns/config/dist.ipfs.io.yaml) (private PL repo). 

We could automate the process and update DNSLink every time `master` branch is updated, 
or have a separate `production` branch if we want to have an additional step for pushing changes to production.

The question: do we want to do this, or is the requirement  for manual PR against private repo a security feature?

cc @aschmahmann @Stebalien @hsanjuan @olizilla @BigLep for any thoughts

Stebalien:
A PR against master makes sense, but we'll want to double check to lock down who can do that. That's how our websites update as well.

lidel:
We had a quick chat about this and the safer design is to automate the  PR against  `protocol/infra` as a safety feature.

### Proposed design

Tweak  `main.yml` workflow when it builds new commit pushed to the master branch (eg. after a PR merge):

1. At the end of `master` build, check if CID changed (compare to current `/ipns/dist.ipfs.io`)
2. If changed,  open PR at  `protocol/infra`  titled "Update DNSLink at dist.ipfs.io" with link to the commit/github action.

Note: this requires token that has PR permissions against   `protocol/infra`, which is bit tricky.

lidel:
After discussing this today, and surprises in https://github.com/protocol/infra/pull/756  we decided to collapse complexity and go with simplified flow:

## Proposed design (B)

Make `dist.ipfs.io` DNSLink update work the same as  `webui.ipfs.io`: update it every time master changes.

1. Move `dist.ipfs.io` management from https://github.com/protocol/infra to dnssimple 
2. add DNSSIMPLE token to this repo
3. Update dnslink when master  is updated  (prior art: [ipfs-webui/.github/workflows/ci.yml#L133-L137](https://github.com/ipfs/ipfs-webui/blob/40a0ce536cb6a9c63b1fdb89f261b96e3fd391b8/.github/workflows/ci.yml#L133-L137))


@petar 1-2 will require help from infra, 3 can be added as the last step  in `persist` stage of [.github/workflows/main.yml](https://github.com/ipfs/distributions/blob/master/.github/workflows/main.yml#L98) in this repo

thattommyhall:
Once the above PR is landed, you will be able to drive changes from here. 
[...]

PR #419, opened 2021-09-30
Update dnslink for dist.ipfs.io on merge to master

petar:
Addressing https://github.com/ipfs/distributions/issues/372.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,18 +1,18 @@
 name: CI
 
-on: push
-  # push:
-  #   branches:
-  #     - master
-  # pull_request:
-  #   branches:
-  #     - master
-  # workflow_dispatch:
-  #   inputs:
-  #     dist_root:
-  #       description: 'DIST_ROOT'
-  #       required: true
-  #       default: '/ipns/dist.ipfs.io'
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+  workflow_dispatch:
+    inputs:
+      dist_root:
+        description: 'DIST_ROOT'
+        required: true
+        default: '/ipns/dist.ipfs.io'
 
 env:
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -1,18 +1,18 @@
...
+  #     - master
+  # pull_request:
+  #   branches:
+  #     - master
+  # workflow_dispatch:
+  #   inputs:
+  #     dist_root:
+  #       description: 'DIST_ROOT'
+  #       required: true
+  #       default: '/ipns/dist.ipfs.io'
```
lidel, line 15: - [ ] Note to self: restore before merge

.github/workflows/main.yml
```diff
@@ -141,6 +145,11 @@ jobs:
         with:
           name: diff
           path: diff
+      # production dnslink is updated on release (during tag build)
+      - run: npx dnslink-dnsimple --domain dist.ipfs.io --link /ipfs/${{ steps.cid-reader.outputs.CID }}
```
lidel, line 149: Before merge, we need to add this, so we update dnslink only when master is built (and not  on PRs and tags)
```suggestion
      - run: npx dnslink-dnsimple --domain dist.ipfs.io --link /ipfs/${{ steps.cid-reader.outputs.CID }}
        if: github.ref == 'refs/heads/master'
```

lidel (commented):
Looks good, thank you :+1: 

Let's wait for `DNSIMPLE_TOKEN` to be set (pinged in https://github.com/protocol/infra/pull/760#pullrequestreview-768265453)  and we should be good to go.

Before merge, we should make changes from comments below.

lidel (approved):
Thank you @petar!

I've tweaked CI to fix unrelated error (details in https://github.com/ipfs/distributions/pull/417#issuecomment-933439634) and made sure we expose secret and run DNSLink update only on `master`.

lidel:
@petar it worked in `main` branch as expected!  :partying_face: 
https://github.com/ipfs/distributions/runs/3792503808#step:14:10

PR #422, opened 2021-10-20
feat: check that linux-amd64 glibc version < 2.32

### LINKED ISSUE #382: Add CI lint job: verify glibc version of linux binaries
Context: https://github.com/ipfs/distributions/issues/348

>  We should also run some script in this repo like https://gist.github.com/fasterthanlime/17e002a8f5e0f189861c to verify that the glibc versions used are below whatever our support threshold is.

guseggert:
for context: https://github.com/ipfs/distributions/issues/348

closes #382


Dockerfile
```diff
@@ -28,6 +28,8 @@ RUN asdf install
 ENV IPFS_PATH=/asdf/.ipfs
 RUN mkdir -p ${IPFS_PATH} && echo "/ip4/127.0.0.1/tcp/5001" > ${IPFS_PATH}/api
 
+RUN go install github.com/guseggert/glibc-check/cmd/glibc-check@v0.0.0-20211020180227-964e8f06bf27 && asdf reshim
```
lidel, line 31: This will work only on CI and `./dockerized` builds.

Can we move this to the end of `./deps-check.sh`  so the build works outside of docker?
(An alternative is to forbid non-dockerized builds by adding a check to ./deps-check.sh` informing user that build can only be run  via `./dockerized` wrapper, but I think it's better to keep the build working everywhere, makes dev and debug easier).

guseggert: Ah yeah fair enough, how about I add a check to `./deps-check.sh`, it would be surprising for that to install stuff...can just leave that up to the dev.

guseggert:
> I am not familiar with the low-level issues around glibc: @aschmahmann do we want this to run only for amd64 or are other archs also impacted? (I imagine it is possible for amd64 and arm64 distro to ship with different glibc versions because of reasons upside_down_face )

linux-arm64 is statically linked since we haven't setup a C cross-compilation toolchain for that (which also means CGo is disabled for arm64, I don't know what the impact is of that?). I checked and Mach executables (macOS) are dynamically linked, so we probably have a similar issue there with libSystem, and likely with Windows too.

Dockerfile
lidel: yes, sgtm

aschmahmann (approved):

PR #443, opened 2021-11-06
fix(ci): more robust cluster connect

lidel:
This PR aims to fix nightly builds: bumps binaries used for build and restores use of bootstrappers, so there is a fallback in case cluster is down or in a weird state.

(CI is green, failures are from previous runs)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,12 +1,12 @@
 name: CI
 
 on:
-  #  push:
-  #  branches:
-  #    - master
-  #pull_request:
-  #  branches:
-  #    - master
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
   workflow_dispatch:
     inputs:
       dist_root:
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -1,12 +1,12 @@
...
-      - master
-  pull_request:
-    branches:
-      - master
+  #  push:
+  #  branches:
+  #    - master
+  #pull_request:
+  #  branches:
+  #    - master
```
lidel, line 9: :information_source: uncomment before merge

scripts/ci/pin-to-cluster.sh
```diff
@@ -11,6 +11,7 @@ ipfs-cluster-ctl --enc=json \
     "$PIN_CID"
 echo "::endgroup::"
 
+# TODO: below can now be replaced with ipfs-cluster-ctl --wait now (https://github.com/ipfs/ipfs-cluster/blob/v0.14.1/CHANGELOG.md#v0141---2021-08-18)
```
lidel, line 14: :thought_balloon:  I am considering leaving it as-is, as it is easier to debug.

lidel (commented):
Merging to fix content routing in nightly.

PR #513, opened 2021-12-09
chore: release fs-repo-11-to-12 v1.0.0

aschmahmann:
A few hiccups to address to it's easier for the next person to do this:

- Manually added in the no-site file since this is a migration, should be a flag in the script or at least in the README
- Manually copied over the repo-owner file, should be from the script or at least in the README
- We should probably call out in the README to use fs-repo-x-to-y/v1.0.0 as the version number when we "enter a new version number"


aschmahmann:
Superseded by #517

PR #520, opened 2021-12-14
chore: use common ipfs/*-actions in setup-ipfs

galargh:
I want to replace the ipfs build and ipfs daemon start logic here with common actions(https://github.com/ipfs/download-ipfs-distribution-action + https://github.com/ipfs/start-ipfs-daemon-action) to minimise duplication. This is part of work related to this issue: https://github.com/ipfs/go-ipfs/issues/8511

The workflow passes after the changes introduced by me: https://github.com/ipfs/distributions/runs/4531714284?check_suite_focus=true

Note that the `download-ipfs-distribution-action` does not check sha512sum yet(https://github.com/ipfs/download-ipfs-distribution-action/issues/3). I'll work on adding it next.


.github/actions/setup-ipfs/action.yml
```diff
@@ -0,0 +1,41 @@
...
+      with:
+        name: ipfs-cluster-ctl
+        version: "${{ env.CLUSTER_CTL_VER }}"
+    - name: Init IPFS daemon
+      run: |
+        # fix resolv - DNS provided by Github is unreliable for DNSLik/dnsaddr
+        sudo sed -i -e 's/nameserver 127.0.0.*/nameserver 1.1.1.1/g' /etc/resolv.conf
+        ipfs init --profile flatfs,server,randomports,lowpower
+        # make flatfs async for faster ci
+        ipfs config --json 'Datastore.Spec.mounts' "$(ipfs config 'Datastore.Spec.mounts' | jq -c '.[0].child.sync=false')"
```
galargh, line 20: We can't init on ipfs daemon start with `--init` flag because we want to modify the config between init and start.

lidel (approved):
thank you! lgtm, removes duplication

PR #534, opened 2021-12-15
fix: update nightly status on commits

lidel:
Problem – https://github.com/ipfs/distributions/issues/531 got fixed, but the "Problem with" message on commit remains: 

>

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -75,13 +75,6 @@ jobs:
               console.log('commented on existing open issue', created)
               link = created.data.html_url
             }
-            await github.repos.createCommitStatus({ ...opts,
-              sha: '${{ github.sha }}',
-              state: 'error',
-              target_url: link,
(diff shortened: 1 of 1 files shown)


galargh (approved):

galargh:
There's definitely a way to update the check instead of creating a new one: https://docs.github.com/en/rest/reference/checks#update-a-check-run

If you think that'd be worth the work here, let me know - I can look into it.

lidel:
@galargh yeah, I think if we can update then it's worth keeping the status – this is low priority – feel free to take over this PR when you have time.

lidel:
Let's merge this for now, to avoid false-negatives in https://github.com/ipfs/distributions/commits/master

galargh:
> @galargh yeah, I think if we can update then it's worth keeping the status – this is low priority – feel free to take over this PR when you have time.

Sorry, I missed this message. I created an issue so that I remember to come back to this - https://github.com/protocol/w3dt-stewards/issues/31

PR #649, opened 2022-03-17
fix: brew tap mitchellh/gon

lidel:
This PR should fix macOS signing with mitchellh/gon by applying a workaround for https://github.com/mitchellh/gon/issues/56

HAhaa


guseggert (approved):

lidel:
Confirmed, the bug got pacified and signing works once again: 
https://github.com/ipfs/distributions/runs/5592009903?check_suite_focus=true#step:8:103

PR #679, opened 2022-04-08
Chore/release go ipfs v0.12.2 v0.11.1

aschmahmann:


dists/go-ipfs/versions
```diff
@@ -60,6 +60,8 @@ v0.10.0
 v0.11.0-rc1
 v0.11.0-rc2
 v0.11.0
+v0.11.1
```
aschmahmann, line 63: I'm not sure if it matters where in this file `v0.11.1` goes, i.e. if it should be in semver order or chronological order.

The tooling uses chronological ordering but I can't tell if that's a bug or not.

lidel: If someone cares, there is `date` field in `dist.json` created during the build, it indicates when release was published.

lidel (approved):

guseggert (approved):

dists/go-ipfs/current
```diff
@@ -1 +1 @@
-v0.12.1
+v0.12.2
```
aschmahmann, line 1: This had to be done manually/carefully to avoid `v0.11.1` being here. Filed for tracking purposes #680

PR #714, opened 2022-05-05
feat: update nightly & signal the ipdx team on failure

laurentsenta:
- Tag the ipdx team when the nightly build fails,
- Skip nightly build for legacy code, use the `ignored-during-nightly` list of files,
- Use go version 1.17 for build, that should fix the current failure with quic-go

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,9 @@ all: deps releases all_dists site
 
 DISTS = $(notdir $(wildcard dists/*))
 
-NDISTS = $(DISTS:%=nightly-%)
+NIGHTLY_IGNORED := $(shell cat ignored-during-nightly)
+DISTS_FILTERED = $(filter-out $(NIGHTLY_IGNORED),$(DISTS))
+NDISTS = $(DISTS_FILTERED:%=nightly-%)
 
 .PHONY: all all_dists deps nightly
 all_dists: $(DISTS)
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -60,7 +60,7 @@ jobs:
             let link
             if (response.data.items.length === 0) {
               const created = await github.issues.create({ ...opts, title, body,
-                labels: ['kind/bug', 'need/triage']
+                labels: ['kind/bug', 'need/triage', 'team/ipdx']
               })
               console.log('no open issues, created a new one', created)
               link = created.data.html_url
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -4,7 +4,9 @@ all: deps releases all_dists site
 
 DISTS = $(notdir $(wildcard dists/*))
 
-NDISTS = $(DISTS:%=nightly-%)
+NIGHTLY_IGNORED := $(shell cat ignored-during-nightly)
+DISTS_FILTERED = $(filter-out $(NIGHTLY_IGNORED),$(DISTS))
+NDISTS = $(DISTS_FILTERED:%=nightly-%)
 
 .PHONY: all all_dists deps nightly
 all_dists: $(DISTS)
```
laurentsenta, line 12: (creating a thread here)
@aschmahmann following up with the failures during nightly builds,

- It looks like nightly is now failing because we use 1.18, go-ipfs & other have not updated quic-go yet => this PR uses 1.17,
- It looks like legacy code builds with 1.18, but figured it make sense to skip legacy code anyway,

aschmahmann: go-ipfs is now building with 1.18 and I'd rather not hold back just because some others haven't yet (e.g. we can get out some PRs quickly to update the relay daemon and ipget, and @hsanjuan will cover the cluster next time he does a release).

In the meanwhile having `ignore-during-nightly` means we can do things like ignore cluster builds and file an issue indicating we should update cluster and remove the ignore.

.tool-versions
```diff
@@ -1,2 +1,2 @@
 nodejs 16.2.0
-golang 1.18.1
```
aschmahmann, line 2: Let's leave this as 1.18.1. My thinking is:
1. Our general policy is to support the last two minor versions of Go (and pretending that X.Y.0 doesn't exist because it's frequently buggy) which means we should be able to build with the latest version.
2. I don't want one project being slow to cause others to be slow

Ideally I think we'd support building with two Go versions (in this case 1.17 and 1.18) and allow some level of configuration around what we build with (e.g. if we really wanted to do a patch release for a binary only built for 1.17 we'd currently be sunk).

cc @hsanjuan for thoughts here

hsanjuan:
> We'll wait for confirmation from @hsanjuan, but we're leaning towards #1

I'm fine with that.

Cluster is getting a release before tuesday and it will be able to build with go1.18 (master already is).

.tool-versions
hsanjuan: agree about leaving golang 1.18.1 here.

PR #717, opened 2022-05-11
feat: dual publishing go-ipfs and kubo

lidel:
## TLDR 

This PR ensures every new release of kubo will be automatically repackaged and  published under the legacy go-ipfs name.

Adding release via  `./dist add-version kubo <version>` will take care of everything.

## Rationale 

We want  the _minimal_ rename (https://github.com/ipfs/go-ipfs/issues/8959) to have a _minimal_ blast radius.

This means we MUST have backward-compatible go-ipfs build still happening, so everything that was created in the past 7+ years and expects 'go-ipfs' will continue working.   

## How this works 

The website will only show "kubo", but Kubo artifacts will be published under both names:

- https://dist.ipfs.io/go-ipfs/
- https://dist.ipfs.io/kubo/

When it comes to adding new Kubo releases:

- We no longer use `go-ipfs` directly,  it is hidden from the website and  `./dist` will error on  `./dist add-version go-ipfs`
- New versions are added via `./dist add-version kubo <version>`
  - This will automatically add corresponding version to legacy go-ipfs, triggering second build next to the official kubo one
  - go-ipfs build no longer compiles go code: instead, it will reuse existing artifacts from kubo and repackage them under the old name and structure

# PREVIEW

> Screen Shot 2022-07-01 at 15 31 15

## TODO

- [x] change ipfs-banana to the final name (proposals in: https://github.com/ipfs/ipfs/discussions/471)
- [x] error on `./dist add-version go-ipfs`
- [x] create go-ipfs release on `./dist add-version kubo`
- [x] always build kubo first, and reuse its artifacts during go-ipfs build (compile only once)
- [x] exclude go-ipfs from nightly builds
- [x] update website (hide go-ipfs, only show the new kubo dist)
- [x] update repository name in both go-ipfs and kubo dists
  - :warning: this must happen after `https://github.com/ipfs/go-ipfs` is renamed to `https://github.com/ipfs/kubo`


aschmahmann:
> and publish both subdirectories on each release:

If we do this it would be great if the data (e.g. the CIDs and SHA-512 checksums) were the same so that it would deduplicate cheaply.

lidel:
FWIW this could be merged as is (creating a retroactive kubo 0.13.0 release) – this way we would not have to wait for  0.13.1 (patch release would happen under both names)

lidel:
This is blocked until we tag  0.14.0-rc1 – then it should be re-run.

Makefile
```diff
@@ -2,7 +2,10 @@ export NODE_OPTIONS="--unhandled-rejections=strict"
 
 all: deps releases all_dists site
 
-DISTS = $(notdir $(wildcard dists/*))
+# DISTS are sorted in reverse lex order
+# because 'kubo' MUST build before 'go-ipfs'
```
Jorropo, line 6: That some magic code I see here. :rofl: 

I don't think it's worth fixing.

PR #725, opened 2022-06-21
feat: cosmetic reame to kubo

lidel:
This PR extracts cosmetic presentation changes from https://github.com/ipfs/distributions/pull/717
and only adds "kubo" to the HTML of dist.ipfs.io website,  
does not change build pipeline, it still uses go-ipfs.

Can be merged without any risk before #717 

- [x] show `kubo (go-ipfs)` but use links to artifacts under https://dist.ipfs.io/go-ipfs/ for now
- [x] bump Nodejs to LTS version
- [x] remove confusing  `gx` and `gx-go`  ([everyone moved to gomod](https://github.com/libp2p/go-libp2p/issues/589))

### Preview

174882784-10f82a45-ad29-44fe-82c7-2d2b5be06f18

cc @aschmahmann as we discussed this during kubo sync today

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,8 +16,8 @@ on:
 
 env:
  DIST_ROOT: ${{ github.event.inputs.custom_dist_root || '/ipns/dist.ipfs.io' }} # content root used for calculating diff to build
- GO_IPFS_VER: 'v0.12.0'          # go-ipfs daemon used for chunking and applying diff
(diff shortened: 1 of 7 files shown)


lidel:
@aschmahmann  updated description in https://github.com/ipfs/distributions/pull/725/commits/f0f4eb7fa2c2a147951cdbf3c10a5dc5d1c8abc6, but it won't be visible yet – string is attached to release, so we will see new description around v0.14 or v0.13.1.

PR #736, opened 2022-07-07
fix(add-version): only run build on CI

Jorropo:


lidel (commented):
(will review after #717  is merged)

Jorropo:
Github's UI isn't actualising and doesn't understand that we merged some commits of this in master already, the real diff is:
```patch
diff --git a/dist.sh b/dist.sh
index 05df1af..3e15a8f 100755
--- a/dist.sh
+++ b/dist.sh
@@ -105,7 +105,9 @@ case $1 in
 
                # cd "dists/$dist" && make update_sources
                # build-go will update sources as needed
-               cd "dists/$dist" && make
+               if [ "$CI" == "true" ]; then
+                       cd "dists/$dist" && make
+               fi
                ;;
        *)
                echo "unrecognized command $1"
```

lidel (approved):

PR #756, opened 2022-09-16
Use go1.19 in the build pipeline

hsanjuan:
I can't release cluster otherwise.


hsanjuan:
@ipfs/go-core-team FYI, maybe revert later if needed

PR #764, opened 2022-09-28
chore: bump publishing deps

lidel:
Mostly addressing cluster bugfix release – details in https://github.com/ipfs-cluster/ipfs-cluster/blob/master/CHANGELOG.md


lidel:
CI did not start at all for some reason, gave it a manual nudge and now running in https://github.com/ipfs/distributions/actions/runs/3145160826

PR #783, opened 2022-12-08
Revert "feat: add support for GOAMD64"

Jorropo:
Reverts ipfs/distributions#782, this is not the right commit, this code doesn't work.


Jorropo:
Supper seeded by https://github.com/ipfs/distributions/pull/784

PR #785, opened 2022-12-08
fix: creating array in build-go script

galargh:
Tested in https://github.com/galargh/.github/actions/runs/3649814822/jobs/6164983080

### DIFF
--- a/build-go.sh
+++ b/build-go.sh
@@ -97,7 +97,7 @@ function goBuild() {
 	(
 		export GOOS="$goos"
 		if [[ "$goarch" == amd64-* ]]; then
-			IFS="-"; read -ar arr <<<"$goarch"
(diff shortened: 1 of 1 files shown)


build-go.sh
```diff
@@ -97,7 +97,7 @@ function goBuild() {
 	(
 		export GOOS="$goos"
 		if [[ "$goarch" == amd64-* ]]; then
-			IFS="-"; read -ar arr <<<"$goarch"
+			IFS="-" read -ra arr <<< "$goarch"
```
Jorropo, line 100: Bash have special sementics when setting IFS before read on the same line it resets IFS, but that more readable so let's keep it that way.

Jorropo (approved):
LGTM

PR #791, opened 2022-12-12
fix: unify build matrix for amd64-v2 and v3

Jorropo:


lidel (changes requested):
Hm.. I realized we've already added this to the template for all other platforms in https://github.com/ipfs/distributions/pull/782  :grin: 

So we either need to merge this to make things consistent (enable here AND in Kubo), or undo https://github.com/ipfs/distributions/pull/782. 

My vote would be to undo: remove `-v2` and `-v3` from template, and only enable it in apps where it makes sense (backed by a benchmark).

@Jorropo what would be good benchmark for Kubo? `test/sharness/t0040-add-and-cat.sh` ? doing many libp2p handshakes (Noise/TLS)?

Jorropo:
@lidel add and cat wont show anything interesting because:
1. hash function use runtime reflection to pick which accelerated implementation to use.
    So enabling -v2 or -v3 wont make the hash function faster (which is the hot code).
2. Add and cat are heavy on IO which again is not changing with that.
3. go-bitswap is very context switch and channel heavy which is pretty indifferent  to micro architecture level.

I think we should run 10 x 10 Kubo nodes, 10 running -v3 10 with -v1 for lets say 3 days. It is critical that no one core is at 100% usage else switching from v1 to v3 will probably still be indicating 100% however v3 should have handled more requests. In other words we must not have the CPU being a bottleneck.
And then see what the average CPU % utilization is. (I would guess that -v3 is slightly lower).

lidel (commented):
Merged https://github.com/ipfs/distributions/pull/792 (which removes -v2 and -v3 from template for non-Windows archs). 

I am also  closing this PR, we can revisit this once we have time to do the benchmark and its results  make it worth the 300% of build time.

PR #856, opened 2023-06-13
Support plugin builds

hsanjuan:
Placing a "plugin" file in the `dists/<repo>` folder will trigger a -buildmode=plugin build using CGO.

### DIFF
--- a/build-go.sh
+++ b/build-go.sh
@@ -94,6 +94,8 @@ function goBuild() {
 	local package="$1"
 	local goos="$2"
 	local goarch="$3"
+	local plugin="$4"
+
 	(
 		export GOOS="$goos"
 		if [[ "$goarch" == amd64-* ]]; then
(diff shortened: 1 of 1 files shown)


build-go.sh
```diff
@@ -211,6 +225,7 @@ function printInitialDistfile() {
   "owner": "$(< repo-owner)",
   "description": "$(< description)",
   "date": "$(date -u '+%B %d, %Y')",
+  "plugin": "$plugin",
```
aschmahmann, line 228: where is this set? (sorry if this is just obvious, my shell is pretty meh)

build-go.sh
```diff
@@ -240,13 +255,19 @@ function buildWithMatrix() {
 	local distname
 	distname=$(basename "$(pwd)")
 
+	local plugin="false"
+	if [ -e plugin ]; then
+	    plugin="true"
+	fi
```
aschmahmann, line 261: Are you manually adding a `plugin` file to indicate something is a plugin? If so could you document it, and if not too painful add it to `dist.sh new-go-dist`? If it's a pain then just documenting should be fine.

aschmahmann (approved):
I assume you've been testing with #857, has that been working alright?

Don't see anything outstanding outside of the two comments below so LGTM, although I haven't done any local testing so if anyone else wants to give some 👀 (especially ones that barely know bash 😅) that could be useful.

hsanjuan:
> I assume you've been testing with #857, has that been working alright?
> 
> Don't see anything outstanding outside of the two comments below so LGTM, although I haven't done any local testing so if anyone else wants to give some eyes (especially ones that barely know bash sweat_smile) that could be useful.

It "works"... but I don't find it very clean, the approach is a bit arbitrary.

Moreover the CI is not prepared to build on anything non linux/amd64, because gcc cross-compiling tools are not installed, and I look to at least have darwin and apparently it is not as easy as installing a gcc package.

So I'm not sure it is worth merging. I thought building the plugin in the exact conditions that Kubo is built would ensure compatibility, but also apparently the plugin just works fine now by building directly by hand (I have no idea if it has to do with Boxo, with re-organizing code, with tagged releases of things or what exactly...).

PR #857, opened 2023-06-13
Nopfs plugin builds (DNM)

hsanjuan:
I am going to abuse the CI here for some builds.


lidel:
My understanding is this is no longer needed since plugin is shipping with  Kubo since https://github.com/ipfs/kubo/pull/10161

hsanjuan:
Sorry I forgot to close it.

Issue #673, opened 2022-03-25
Failed binary builds are seen as successful


hsanjuan:
```
==> building for openbsd amd64
    output to ../../releases/ipfs-cluster-service/v1.0.0-rc1/ipfs-cluster-service_v1.0.0-rc1_openbsd-amd64
# github.com/elastic/gosigar
../gopath/pkg/mod/github.com/elastic/gosigar@v0.12.0/concrete_sigar.go:20:11: cpuUsage.Get undefined (type Cpu has no field or method Get)
../gopath/pkg/mod/github.com/elastic/gosigar@v0.12.0/concrete_sigar.go:30:13: cpuUsage.Get undefined (type Cpu has no field or method Get)
../gopath/pkg/mod/github.com/elastic/gosigar@v0.12.0/concrete_sigar.go:49:10: l.Get undefined (type LoadAverage has no field or method Get)
../gopath/pkg/mod/github.com/elastic/gosigar@v0.12.0/concrete_sigar.go:55:10: m.Get undefined (type Mem has no field or method Get)
../gopath/pkg/mod/github.com/elastic/gosigar@v0.12.0/concrete_sigar.go:61:10: s.Get undefined (type Swap has no field or method Get)
../gopath/pkg/mod/github.com/elastic/gosigar@v0.12.0/concrete_sigar.go:67:10: p.Get undefined (type HugeTLBPages has no field or method Get)
../gopath/pkg/mod/github.com/elastic/gosigar@v0.12.0/concrete_sigar.go:73:10: f.Get undefined (type FileSystemUsage has no field or method Get)
../gopath/pkg/mod/github.com/elastic/gosigar@v0.12.0/concrete_sigar.go:79:11: fd.Get undefined (type FDUsage has no field or method Get)
../gopath/pkg/mod/github.com/elastic/gosigar@v0.12.0/concrete_sigar.go:87:10: r.Get undefined (type Rusage has no field or method Get)
    openbsd amd64 build succeeded!
```

hsanjuan:
Also:

```
==> building for freebsd 386
    output to ../../releases/ipfs-cluster-service/v1.0.5/ipfs-cluster-service_v1.0.5_freebsd-386
# github.com/cockroachdb/pebble/internal/batchskl
../gopath/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20221122204154-936e011bb911/internal/batchskl/skl.go:310:18: maxNodesSize (untyped int constant 4294967295) overflows int
[...]

PR #1067, opened 2024-01-31
fix: correct go build error passing

lidel:
Aims to Close https://github.com/ipfs/distributions/issues/673

The meta reason is the entire build pipeline not being written with `set -e`, which means every potential error path needs to be manually handled, or it will be ignored.

For fix here, I suspect things worked fine until we added `glibc-check` here, which made `go build` no longer be the last command in `goBuild`, and if `go build` fails its code is not the code returned by `goBuild`, but we continue to the next command, which may run ok due to no binaries to check :upside_down_face: 

## TODO

- [x] fix the way build-log is created and go build errors are handled
- [x] try to simulate failure on CI and confirm error is handled ok 
- [x] switch back DIST_ROOT to default and merge if passing

### DIFF
--- a/build-go.sh
+++ b/build-go.sh
@@ -106,9 +106,10 @@ function goBuild() {
 
 		local output
 		output="$(pwd)/$(basename "$package")$(go env GOEXE)"
-		go build -mod=mod -o "$output" \
-			-trimpath \
-			"${package}"
+		if ! (go build -mod=mod -o "$output" -trimpath "${package}"); then
+			warn "    go build with $package $goos $goarch failed."
+			return 1
+		fi
 
 		if [ -x "$(which glibc-check)" ] && [ "$GOOS" == "linux" ] && [ "$GOARCH" == "amd64" ]; then
 			echo "GLIBC versions:"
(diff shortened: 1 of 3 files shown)


build-go.sh
```diff
@@ -106,9 +106,10 @@ function goBuild() {
 
 		local output
 		output="$(pwd)/$(basename "$package")$(go env GOEXE)"
-		go build -mod=mod -o "$output" \
-			-trimpath \
-			"${package}"
+		if ! (go build -mod=mod -o "$output" -trimpath "${package}"); then
+			warn "    go build of $output failed."
+			return 1
+		fi
```
lidel, line 112: :information_source:  error handling here was missing

build-go.sh
```diff
@@ -144,7 +145,7 @@ function doBuild() {
 
 	mkdir -p "$dir"
 
-	if ! (cd "$build_dir_name" && goBuild "$package" "$goos" "$goarch") > build-log; then
+	if ! (cd "$build_dir_name" && goBuild "$package" "$goos" "$goarch") > "$build_dir_name/build-log" 2>&1; then
```
lidel, line 148: :information_source:   why error logs were not created correctly OR were missing stderr

lidel (commented):
Waiting for CI to pass, if not issues this is ok to merge.

PR #1108, opened 2024-11-08
fix: do not include full path to binary in .sha512 files

hsanjuan:
See https://github.com/ipfs/kubo/issues/9323.

This is a darwin release thing because the signature process for macOS means that we recalculate sha512 and cid and we do this using an absolute path so the result includes the path to the file.

The fix is to cd to the folder and do it from there.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ on:
         default: '/ipns/dist.ipfs.tech'
 
 env:
- DIST_ROOT: ${{ github.event.inputs.custom_dist_root || '/ipns/dist.ipfs.tech' }} # content root used for calculating diff to build
+ DIST_ROOT: ${{ github.event.inputs.dist_root || '/ipns/dist.ipfs.tech' }} # content root used for calculating diff to build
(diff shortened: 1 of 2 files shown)


lidel:
Thanks! Manually dispatched https://github.com/ipfs/distributions/actions/workflows/main.yml
with IPFS_DIST=/ipfs/QmeqkkydCxmREqg2qmoA7XQSYbcgvSko5zDAhnarKFp2Ki (version without 0.32.0-rc2) to see if produced DAG has expected values.

lidel (approved):
lgtm, https://github.com/ipfs/distributions/actions/runs/11750043494/job/32738190122#step:9:10 produced DAG with expected value:

```
$ ipfs cat /ipfs/Qmbu8BBcVVT6JpnQWssMwmzGaD6YQeB8XV7PtU4yp1BgNo/kubo/v0.32.0-rc2/kubo_v0.32.0-rc2_darwin-amd64.tar.gz.sha512
dccbd57d1d66e9cb95845c9b251d4b190771141451f50bb4e88710f4e4d869356a916de0f5ef14ec23f633865c88451194a2565036fac929bd2339424cde167b  kubo_v0.32.0-rc2_darwin-amd64.tar.gz
```

PR #1124, opened 2025-01-24
chore(ci): restore download-ipfs-distribution-action

lidel:
Testing upstream version with fix from https://github.com/ipfs/download-ipfs-distribution-action/pull/37


lidel:
Looks good:
- https://github.com/ipfs/distributions/actions/runs/12953678607/job/36133900287?pr=1124#step:4:7

PR #1149, opened 2025-07-08
feat: kubo v0.36.0 rc2 & go 1.24.5

lidel:
- https://github.com/ipfs/kubo/issues/10816#issuecomment-3050108416


lidel:
CI failing sue to apple paperwork, reached out to relevant people to check the checkbox again (cc https://github.com/ipfs/ipfs-desktop/issues/1753)

lidel:
since this is RC, we don't want to block on Apple Signing. 

i've added code that allows us skipping it for cases like this RC, hopefully this will be resolved before final release next week (we do want final release to be signed and notarized)

PR #1172, opened 2025-10-02
feat: add RISC-V architecture support

lidel:
- add linux riscv64 to build matrices for kubo and ipfs-cluster projects
  - cc @hsanjuan  if any concern with adding prebuilt cluster binaries for RISC-V?  -- should be fine since we use >1.25 now https://go.dev/doc/go1.25 and I've been running Kubo 0.38 on RISC-V just fine
- add riscv64 to website targetMap for download table display
- update architecture labels to more intuitive industry-standard naming:
  - amd64: "AMD64"
  - arm64: "ARM64"
  - riscv64: "RISC-V"


lidel:
Assuming no concerns – I will merge once CI is green.

